PR #4, opened 2022-10-21
Front end changes to allow editing task data

jbirddog:
The edit option is currently only enabled for tasks in the ready state. Hitting edit will load the task data in a monaco editor like edit json files. On save calls an api which @cullerton has for the backend. Cancel takes you back to readonly mode and drops any changes that were not saved.

Collab with @cullerton


spiffworkflow-frontend/src/routes/ProcessInstanceShow.tsx
```diff
@@ -211,29 +225,125 @@ export default function ProcessInstanceShow() {
...
+  const saveTaskDataResult = (_: any) => {
+    setEditingTaskData(false);
+    const dataObject = taskDataStringToObject(taskDataToDisplay);
+    const taskToDisplayCopy = { ...taskToDisplay, data: dataObject }; // spread operator
+    setTaskToDisplay(taskToDisplayCopy);
+  };
+
+  const saveTaskDataFailure = (result: any) => {
+    console.log(result.toString());
+    // TODO: Not sure what to do here
```
burnettk, line 250: you could potentially just display a message to the user on failure using setErrorMessage, which is used in various components.

burnettk (approved):

PR #5, opened 2022-10-24
XML Validation Fixes

danfunk:
Based off validation errors we encountered in Spiff Workflow, these fixes should assure we generate XML that will validate against he officially published BPMN schemas.

* For Correlation Properties, should use messagePath and not formalEx…pression for the name.
* When all data input and outputs are removed, the ioSpecification should be removed as well
* If an ioSpecification is present, then there shoud be a at least one dataInputset and at least one dataoutputset.

### DIFF
--- a/bpmn-js-spiffworkflow/app/spiffworkflow/InputOutput/IoInterceptor.js
+++ b/bpmn-js-spiffworkflow/app/spiffworkflow/InputOutput/IoInterceptor.js
@@ -61,7 +61,6 @@ export default class IoInterceptor extends CommandInterceptor {
         if (context.shape.di.$parent) {
           collectionRemove(context.shape.di.$parent.planeElement, context.shape.di);
         }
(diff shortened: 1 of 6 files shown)


bpmn-js-spiffworkflow/app/spiffworkflow/InputOutput/IoInterceptor.js
```diff
@@ -60,6 +61,10 @@ export default class IoInterceptor extends CommandInterceptor {
         if (context.shape.di.$parent) {
           collectionRemove(context.shape.di.$parent.planeElement, context.shape.di);
         }
+        console.log("inputs/outputs", ioSpec.dataInputs, ioSpec.dataOutputs)
```
burnettk, line 64: should we kill the console?

burnettk (approved):

essweine (approved):

PR #15, opened 2022-11-02
From the logs, allow viewing a diagram in a previous state

jbirddog:
Collab with @essweine 

Associates a `spiff_step` with each iteration of SpiffWorkflow for a given process instance. For each step the current state of the workflow is serialized. When the logs are returned to the front end the associated step is included. When viewing a process instance this step can optionally be included to view the instance as it was at that point in time.

Due to the use of `do_engine` steps in the backend there is not a lot of granularity when viewing the process instance at a previous time. This will be addressed in work slated for this month. The easiest way to see this feature currently is to have a user/manual task that will break up the spiff flow into multiple steps.


burnettk (approved):
did you consider not having a "spiff_step" increasing integer but just the spiff step details records?

jbirddog:
> did you consider not having a "spiff_step" increasing integer but just the spiff step details records?

I was hoping not to have the incrementing id honestly, but couldn't see a way around not saving the serialized workflow/tasks per spiff log. Each log entry from spiff needs some ahead of time value to associate with the incoming log when the filter runs. And multiple log lines from spiff can be associate with each "step". Maybe when we move on from `do_engine_steps` and manually walk the tasks this can be done. Honestly not 100% sure.

burnettk:
oh, so it's like you need to have a value before the spiff step detail record exists, eh. cool, thanks. maybe we just keep an eye on it and see if there's an opportunity to kill it in the future, but no biggie.

PR #23, opened 2022-11-08
First pass at custom report/perspective for Process Instance List

jbirddog:
First pass at integrating the existing reporting subsystem with the process instance list view. The main model level change is that reports are no longer associated with a process group/model, but instead with the user that created them. When the process instance list is returned from the backend it attempts to load a report called "default" for the current user. If one does not exist the system default is returned. The frontend then uses the logic from the existing report screen to generate the process instance list. I added some system level formatting logic so instance ids and group/model identifiers could be links like they are today.

I did not change any of the frontend routes/links, will do that in a follow up pr. Also need to look some more into how filtering from the process instance list works when you have a perspective defined, etc. The way a user defined their default report will need some love I suspect. Also might want to mass rename report -> perspective.

Using the system default:
Screenshot from 2022-11-07 15-08-25

User defined defaults:
Screenshot from 2022-11-07 15-05-29


spiffworkflow-backend/tests/spiffworkflow_backend/integration/test_process_api.py
```diff
@@ -1168,6 +1168,8 @@ def test_process_instance_list_with_default_list(
         assert response.json["pagination"]["pages"] == 1
         assert response.json["pagination"]["total"] == 1
 
+        print(response.json)
```
burnettk, line 1171: maybe kill

burnettk (approved):

spiffworkflow-backend/tests/spiffworkflow_backend/integration/test_process_api.py
jbirddog: Yes, thanks.

PR #26, opened 2022-11-08
Bulk insert spiff logs and step details

jbirddog:
Part of the performance improvements from the demo database. Locally this brings submitting the first user form from the mvp flow from 19s to 13s. Strategy is to use `bulk_insert_mappings` which lets you bypass a lot of the ORM magic to perform inserts that do not need to be tracked in the session for relations/dirty updates, etc. Instead of inserting and committing for each step/log entry we build them up and buik insert them along the way. Still need to look into the serialization but this is a step.


jbirddog:
Test failures... drafting until they are fixed.

burnettk:
documented this profiling technique at https://github.com/sartography/onboarding/commit/2d7c15f0cf217a781656460ad84d4d2774bd6b5b

burnettk:
this is hugely valuable investigation and remediation work, thank you. it occurred to me that it might be worth shoving in python 3.11.x as they are claiming some pretty large speed improvements. not sure if much of this is python-limited, but maybe we get a half second or something.

jbirddog:
> this is hugely valuable investigation and remediation work, thank you. it occurred to me that it might be worth shoving in python 3.11.x as they are claiming some pretty large speed improvements. not sure if much of this is python-limited, but maybe we get a half second or something.

Good idea. The next piece of this - the serialization time - i believe is tied up heavily in python execution speed. I'll open a separate pr to bump to 3.11.x in case we need to roll one of these changes back.

jbirddog:
Updated timings after merging in main and latest process models:

main:
Screenshot from 2022-11-09 14-12-51

bulk_insert:
Screenshot from 2022-11-09 14-14-19

burnettk (approved):

PR #27, opened 2022-11-09
Move to Python 3.11

jbirddog:
Needed to bump RestrictedPython and greenlet. Also decent amount of code changes neeeded to make ./bin/run_pyl pass. Also bumped .tool_versions.

Not sure what I need to do about the first line in the backend's Dockerfile, or any other deployment related things to use 3.11.


jbirddog:
SpiffWorkflow's test suite runs ~15% faster with 3.11 vs 3.10.4, so does look like there are some gains.

spiffworkflow-backend/src/spiffworkflow_backend/services/process_model_service.py
```diff
@@ -316,5 +322,5 @@ def __scan_spec(
             with open(spec_path, "w") as wf_json:
                 json.dump(self.WF_SCHEMA.dump(spec), wf_json, indent=4)
         if process_group:
-            spec.process_group_id = process_group.id
+            spec.process_group = process_group.id
```
burnettk, line 325: is this intended?

spiffworkflow-backend/src/spiffworkflow_backend/services/spec_file_service.py
```diff
@@ -481,7 +487,7 @@ def check_for_message_models(
                         ):
                             raise ValidationException(
                                 "Message model is already used to start process model"
-                                f"'{process_model_info.process_group_id}/{process_model_info.id}'"
+                                f"'{process_model_info.process_group}/{process_model_info.id}'"
```
burnettk, line 490: this should probably become just `process_model_info.id` instead of the f string

burnettk (approved):
regarding the Dockerfile thing, `ghcr.io/sartography/python:3.11` is available now since i updated this repo: https://github.com/sartography/docker-python/actions/runs/3430666028

spiffworkflow-backend/src/spiffworkflow_backend/services/process_model_service.py
jbirddog: There was a lint about it when running pyl, believe it looked right at the time - if you don't think it is ok with reverting and noqa'ing it or whatever other option is best.

PR #32, opened 2022-11-14
Feature/call activity selection

danfunk:
Still working on the front end, but this contains some refactoring work to move most of the XML parsing into SpiffWorkflow, and it's touching enough that I don't want to battle a ton of merge conflicts as we work today.


spiffworkflow-backend/src/spiffworkflow_backend/models/message_triggerable_process_model.py
```diff
@@ -16,6 +16,7 @@ class MessageTriggerableProcessModel(SpiffworkflowBaseDBModel):
         ForeignKey(MessageModel.id), nullable=False, unique=True
     )
     process_model_identifier: str = db.Column(db.String(50), nullable=False, index=True)
+    # fixme:  Maybe we don't need this anymore?
```
burnettk, line 19: true

burnettk (approved):

jasquat (approved):

PR #36, opened 2022-11-15
Feature/call activity selection

danfunk:
* Bug in the BPMN-JS -- needed to pass along the event bus
* Renamed FileReference to SpecReference, and BpmnProcessIdLookup to SpecReferenceCache -- because they are the same thing, and that's the best name I could come up with that would pull all the concepts together.  Happy to use a better name, I don't love it.
* ALL processes (multiple processes in a single file, processes that are not primary) are included in the SpecRefernece (ye old BpmnProcessIdLookup)  SpecReference also includes all DMN Decision Ids as well.
* Added an API Endoint

** NOT INCLUDED:  Building the cache when the system starts up.  But would love to talk about it when you have a minute.


burnettk:
can we rename SpecReference (and cache) to ProcessModelReference?

spiffworkflow-backend/migrations/env.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import with_statement
+
```
jbirddog, line 2: I may be mixing this up, but I believe this is added/removed depending on if you run the backend lints from within spiffworkflow-backend or `./bin/run_pyl` from the monorepo root. If that is a bad memory on my end please disregard.

burnettk (approved):

spiffworkflow-backend/migrations/env.py
jasquat: Yeah, KB just ran it and it was removed.

burnettk: true, committed result of run_pyl just now

spiffworkflow-backend/src/spiffworkflow_backend/models/spec_reference.py
```diff
@@ -0,0 +1,84 @@
...
+            type=ref.type,
+            file_name=ref.file_name,
+            has_lanes=ref.has_lanes,
+            is_executable=ref.is_executable,
+            is_primary=ref.is_primary,
+            relative_path=ref.relative_path,
+        )
+
+
+class SpecReferenceSchema(Schema):
```
jasquat, line 66: Since SpecReference is a dataclass you may not need marshmallow here since it'll serialize to json as is. If you need to remove fields you can add a serialized method like for ProcessInstanceModel.

spiffworkflow-backend/src/spiffworkflow_backend/models/spec_reference.py
```diff
@@ -0,0 +1,84 @@
...
+    relative_path: str  # The path to the file.
+    has_lanes: bool  # If this is a process, whether it has lanes or not.
+    is_executable: bool  # Whether this process or decision is designated as executable.
+    is_primary: bool  # Whether this is the primary process of a process model
+    messages: dict  # Any messages defined in the same file where this process is defined.
+    correlations: dict  # Any correlations defined in the same file with this process.
+    start_messages: list  # The names of any messages that would start this process.
+
+
+class SpecReferenceCache(SpiffworkflowBaseDBModel):
```
jasquat, line 34: Add Model to the end of this since that's what we've been doing for all db model classes.

PR #37, opened 2022-11-15
Merge process_instance_list query filters with report filters

jbirddog:
This is a much more straightforward approach that what I was trying to do yesterday. If there is a default report, parse its `filter_by` and overlay the individual filter query params passed into `process_instance_list`. This also means report filtering is done in the db the same was as if you entered a filter manually in the UI. This does not currently support filtering on task data. I also dropped handling of the filter comparison since the `process_instance_list` query filters have an assumed comparison operator.

Next I'd like to serialize the combined filters and send that in the response so the UI can populate the filters appropriately if a report is in play.


jasquat (approved):

burnettk:
hot

PR #44, opened 2022-11-16
The actual working Call Activity search 

danfunk:
Searching for call activities seems to be working now.
I had the clear_caches in the wrong place previously - fixing.

### DIFF
--- a/spiffworkflow-frontend/src/components/ProcessSearch.tsx
+++ b/spiffworkflow-frontend/src/components/ProcessSearch.tsx
@@ -17,14 +17,17 @@ export default function ProcessSearch({
   processes,
   selectedItem,
   onChange,
-  titleText = 'Process model',
+  titleText = 'Process Search',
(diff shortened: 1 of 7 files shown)


spiffworkflow-frontend/src/components/ProcessSearch.tsx
```diff
@@ -17,14 +17,17 @@ export default function ProcessSearch({
   processes,
   selectedItem,
   onChange,
-  titleText = 'Process model',
+  titleText = 'Process Search',
```
burnettk, line 20: i think this also gets used where you're selecting the model on the filter page, so Process model makes more sense there, but probably not a dealbreaker.

burnettk: i take it back, i guess this is a new component.

burnettk (approved):

PR #61, opened 2022-11-23
Fix the postgres job

jbirddog:
Postgres didn't like a distinct, since it did not know how to distinct json columns. Tried forcing a `distinct on` but that results in a warning that a future version will issue a CompilerError for non postgres backends. Using a group_by works and the results appear to be the same as before.


burnettk (approved):
this solution makes more sense than the thing we first found anyway, thanks.

PR #66, opened 2022-12-01
A Demo Docker-Compose for All

danfunk:
1. **Demo Connector Proxy**, so we don't have any weird dependencies on Status Specific connectors.
2. **Git Action for Tags** that will create docker images when you push a tag up - creating three release images - frontend, backend, and proxy.
3. **OpenID Blueprint** that creates an insanely insecure built-in open id client that you can use for demos and local installations.  Provides a log-in screen, and responds correctly to the handful of openID connections we use.
4. **OpenID's well-known/configuration** We now use this well-known endpoint so that our open-id client is more agnostic about implementations.  Looking forward to trying Google.
5. **New Permissions Section**  You can add users to the same permissions file where groups and permissions are defined.  This connects into the openID blueprint.
6. Changes to the order in which we read configuration files - so that instance/config.py is read at the very end, and doesn't get overridden.


spiffworkflow-backend/src/spiffworkflow_backend/config/permissions/demo.yml
```diff
@@ -0,0 +1,88 @@
+default_group: everybody
```
burnettk, line 1: if we call this demo.yml, i think it's going to override for https://demo.spiffworkflow.org. we would ideally make a new environment like https://status.dev.spiffworkflow.org (and https://status.staging.spiffworkflow.org).

spiffworkflow-backend/tests/spiffworkflow_backend/integration/test_openid_blueprint.py
```diff
@@ -0,0 +1,61 @@
+"""Test_authentication."""
+from flask import Flask
+from flask.testing import FlaskClient
+from tests.spiffworkflow_backend.helpers.base_test import BaseTest
+
+
+class TestFaskOpenId(BaseTest):
```
burnettk, line 7: Fask typo

burnettk (approved):
sweet!

spiffworkflow-backend/src/spiffworkflow_backend/models/user.py
```diff
@@ -30,7 +30,8 @@ class UserModel(SpiffworkflowBaseDBModel):
     __table_args__ = (db.UniqueConstraint("service", "service_id", name="service_key"),)
 
     id = db.Column(db.Integer, primary_key=True)
-    username = db.Column(db.String(255), nullable=False, unique=True)
+    # server and service id must be unique, not username.
```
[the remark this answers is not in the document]
danfunk: I ran into a problem where I switched from "dan" as defined in KeyCloak to "dan" who is using the internal demo open-id system - and that created a conflict.  When I looked at the user model, I saw that there is a unique constraint on service and service_id --- so I assumed that the purpose of that was to allow multiple user names to be the same -- and "lalalala" fixed it and moved on.  But I can see where having a unique username in the system could be critical when we are assigning users to groups based on their user name.  So agree -- we should talk about this and decide which direction we want to go.

spiffworkflow-backend/src/spiffworkflow_backend/config/permissions/demo.yml
danfunk: renamed it to example.yml for now to avoid conflicts, and we can deal with new deployment configurations when the time comes.

PR #76, opened 2022-12-14
Allow viewing/editing xml of bpmn and dmn files

jbirddog:
For line 10 in the december spreadsheet - adds support for editing the xml of bpmn and dmn files. Not sure how far we want to take this one - right now you can only view the xml of a file that has been saved. Any changes to the xml need to be saved before switching between editing modes. The work to prompt on save should make this more obvious.


burnettk (approved):
I wonder if "Edit XML" and "Edit Diagram" would be better for the button labels? I think it's fine the way it is. I'm not sure what the primary use case is, editing or viewing. Probably viewing.

jasquat (approved):

PR #77, opened 2022-12-14
Fix endpoints for script task unit tests

jbirddog:
Ran into this when working with @danfunk - this updates the endpoints for the script task unit tests to use `modified_process_model_identifier`.


burnettk:
thanks. jason and i have been mentioning to each other that we should fix this every time we see the feature while working on something else.

burnettk (approved):

jasquat (approved):

PR #78, opened 2022-12-15
Feature/process navigation

essweine:
This https://github.com/sartography/SpiffWorkflow/pull/273 needs to be merged before this.


jbirddog (approved):

burnettk (approved):

burnettk:
is this the sort of thing that should require elevated permissions, @essweine ? or, like, if you kicked off the process instance, then you should be able to send messages to these sorts of activities?

essweine:
I think elevated permissions are probably a good idea -- you could potentially screw something up.  I would say they should be the same as for editing task data.

essweine:
The frontend tests seem to be failing because of timeouts, problems with keycloak, etc.  I don't understand why the backend tests fail on reordering imports because nothing changes when I run it.

burnettk:
please ignore the frontend tests for now. they are currently broken, and
we'll fix them when we have some time. re: imports, are you running
./bin/run_pyl from spiff-arena ?

essweine:
I'm done with this, except for the pre commit script failing.  If I run the linting script it changes almost every file in the repo and then errors out.  If I run only orerdering imports, nothing changes.  It irritates me that stupid stuff like this causes problems, yet we have multi thousand line files with unused and commented out code..  So anyway, maybe somebody else can pull this and run the script locally; I'm officially giving up on it.

burnettk:
sure, hammered `./bin/run_pyl` from spiff-arena and it made some minor changes to two files. you may or may not be happy to learn that we just split up the process_api_blueprint.py file into five or six files based on the "resource type" (process model, process group, task, etc). also, you should merge this before we get that branch into main. but yeah, there's garbage code, for sure. sonarcloud calls us out on commented out code, too. it would be nice to have more time to refactor and clean stuff up. all in all it's decent, though, i think.

PR #91, opened 2023-01-03
Helper function to support reaping process instances

jbirddog:
To be used by a process model that identifies names/status/last_updated_delta for process instances to be reaped. Process instances matching the given criteria are removed along with their associated spiff step details. The example process model can be found [here](https://github.com/sartography/sample-process-models/pull/7).

Also exposed `map` to script tasks, which the sample reaper process model uses.


spiffworkflow-backend/src/spiffworkflow_backend/scripts/delete_process_instances_with_criteria.py
```diff
@@ -0,0 +1,61 @@
...
+            delete_criteria.append(
+                (ProcessInstanceModel.process_model_identifier == criteria["name"])
+                & ProcessInstanceModel.status.in_(criteria["status"])  # type: ignore
+                & (
+                    ProcessInstanceModel.updated_at_in_seconds
+                    < (delete_time - criteria["last_updated_delta"])
+                )
+            )
+
+        results = ProcessInstanceModel.query.filter(or_(*delete_criteria)).all()
```
burnettk, line 45: i guess this will scale to like thousands but not millions, maybe? i guess if we charge $2 / instance, we'll be able to update this then. :)

burnettk (approved):

spiffworkflow-backend/src/spiffworkflow_backend/scripts/delete_process_instances_with_criteria.py
jbirddog: My sqlalchemy foo wasn't strong enough to just issue a delete, seemed like when I did i was below the orm and deletes didn't cascade properly. Thought about doing it all manually but then figured this would be something that always breaks when new relationships are deleted. Definitely open to suggestions on the sql part.

Also just remembered early on we talked about only doing 100 or so instances in a single pass so the system doesn't choke. Can add that in as well.

burnettk: oh, yeah, actually limiting to 100 at a time probably pretty much solves it, nice. i was also thinking along the lines of avoiding the ORM and issuing pure deletes to the db, but it probably isn't necessary with the limit, and yeah, certainly more maintainable if we can let the ORM help us with cascading.

jbirddog: Ok cool, added the 100 limit so will merge - might help reproduce the db growth on dev, who knows. Also open to further improvements.

PR #105, opened 2023-01-17
do not write to logs when a task is inheriting data from the parent w…

jasquat:
…/ burnettk

### DIFF
--- a/SpiffWorkflow/SpiffWorkflow/task.py
+++ b/SpiffWorkflow/SpiffWorkflow/task.py
@@ -708,18 +708,22 @@ class Task(object,  metaclass=DeprecatedMetaTask):
     def _get_internal_data(self, name, default=None):
         return self.internal_data.get(name, default)
 
-    def set_data(self, **kwargs):
+    def set_data(self, write_to_log=True, **kwargs):
         """
         Defines the given attribute/value pairs.
         """
         self.data.update(kwargs)
-        data_log.info('Set data', extra=self.log_info())
+
+        if write_to_log:
+            data_log.info('Set data', extra=self.log_info())
 
     def _inherit_data(self):
         """
         Inherits the data from the parent.
         """
-        self.set_data(**self.parent.data)
+
+        # no need to write to a log when we are just inheriting data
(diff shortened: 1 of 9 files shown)


SpiffWorkflow/SpiffWorkflow/task.py
```diff
@@ -708,18 +708,22 @@ def _set_internal_data(self, **kwargs):
     def _get_internal_data(self, name, default=None):
         return self.internal_data.get(name, default)
 
-    def set_data(self, **kwargs):
+    def set_data(self, write_to_log=True, **kwargs):
```
essweine, line 711: I really don't think we should add this to spiff.  This is exactly the kind of thing I'm trying to remove whenever I see it.  If you don't want to see these log entries, it would be better to filter them on the logging side.

burnettk: that makes sense. how would we differentiate these set data log entries from other set data log entries in order to filter? we were sort of hoping we could avoid an ever-increasing log table when not much is actually happening from the user's perspective. in this case, it seems like spiff is logging "set data" whenever it applies inherited data to the task, which it does every time do_engine_steps runs, which sort of feels like an implementation detail. we also get "set data" log entries when data is changing, and those could be useful, but we would not be able to filter one without filtering the other with the data we get from spiff at the moment.

essweine: Those statements go to the `spiff.data` logger, which was an ill-conceived mistake.  The only cases where it's providing any useful information is when things are written to and from data objects.  Everything else is always done in the context of a task update, so these logs really are redundant.

Additionally, it doesn't even account for script tasks, which is probably where most data changes are taking place.

So instead of making particular log statement optional, I would approve just deleting it all together in this context.

jasquat: Removing the log statement altogether seems fine to me if it's redundant.

burnettk: updated

burnettk (approved):

PR #110, opened 2023-01-25
Feature/jinja errors

danfunk:
Errors here are sacrosanct.  If you feel it in no other way, experience it through the pain I experienced in creating this "humble" pull request.   If Alex or Sasha can't understand the error and act on it, then we have walked them down a dark alley, taken their wallet, sucker punched them in the stomach, and gone to eat a cheeseburger on their dime. 

1) Lots of fixes around the useEffect in TaskShow so that we can display error messages properly.  Mainly it's about deps.
2) Surface better errors when hitting Jinja2 exceptions.
3) Refactored the error context in the front end, which in hindsight was not necessary, but I do think it is a bit cleaner.


spiffworkflow-frontend/src/routes/TaskShow.tsx
```diff
@@ -39,48 +39,55 @@ export default function TaskShow() {
...
       HttpService.makeCallToBackend({
         path: `/tasks/${params.process_instance_id}/${params.task_id}`,
         successCallback: processResult,
-        // This causes the page to continuously reload
-        // failureCallback: setErrorObject,
+        failureCallback: addError,
       });
     }
-  }, [params, permissionsLoaded, ability, targetUris]);
+  }, [permissionsLoaded, ability]); // params and targetUris (which deps on params) cause this to re-fire in an infinite loop on error.
```
jasquat, line 70: I don't think the issue is with the params and targetUris. It may have to do with addError though.

spiffworkflow-frontend/src/routes/TaskShow.tsx
```diff
@@ -39,48 +39,55 @@ export default function TaskShow() {
...
+        failureCallback: (error: any) => {
+          addError(error);
+        },
+      });
+    }
+  };
+
   useEffect(() => {
     if (permissionsLoaded) {
-      const processResult = (result: any) => {
```
jasquat, line 44: Why move this method out of the useEffect? useEffect generally doesn't like it when you move methods outside of it.

burnettk:
hey, dan. we made a few changes to the frontend stuff. we're not totally sure it won't break one of your use cases, but we're gonna merge and shove this out, and we can always discuss and make updates.

PR #114, opened 2023-01-30
frontend: use /api subpath instead of subdomain

jakubgs:
Because buying certificates with multiple SANs every time you want to deploy an envrionment is a pain the ass and expensive.


jakubgs:
The page seems to load fine and at least OpenID works fine. Not sure what else needs to be done, but normally this should be enough, unless you're doing something weird i your frontend codebase.

jakubgs:
Thanks for babysitting this.

burnettk:
thanks for raising this. there are very likely other users who will appreciate having this option available.

PR #116, opened 2023-01-30
frontend: avoid redundant steps in Dockerfile

jakubgs:
Use separate base, setup and final to avoid redundat steps.
Also add default value for `PORT0` as `7001`.


danfunk (approved):

burnettk (approved):
excellent

PR #141, opened 2023-02-20
backend/git_service: check repo URLs from webhook

jakubgs:
Since we are cloning a private repo we are using `ssh_url` in our case.

Also fix handling test requests sent at webhook creation time.
https://docs.github.com/en/webhooks-and-events/webhooks/testing-webhooks


burnettk (approved):
great, thanks

jasquat (approved):

PR #142, opened 2023-02-20
Feature/message fixes

danfunk:
# SpiffWorkflow Library
* SpiffWorkflow event_definitions wanted to return a message event's correlation properties mested within correlation keys. But messages are directly related to properties, not to keys - and it forced a number of conversions that made for tricky code. So Messages now contain a dictionary of correlation properties only.
* SpiffWorkflow did not serialize correlations - so they were lost between save and retrieve.
* When comparing Correlation Property values we were getting negatives due to data types ('1001' != 1001)  - we are storing these values as strings in the database and can't convert them back to integers later, so I changed everything everywhere to compare after conversion to a string.  (Don't feel great about this one.)
* Create decent error messages when correlations fail

# SpiffWorkflow Backend: Messages
* Link between message instance and correlations is now a link table and many-to-many relationships in the 'SqlAlchemy Way', which helps remove a lot of boilerplate code.  (removed get_message_instance_receive among several others). Required a bit of a weird database migration.
* Match the correlation properties, not the process id when accepting api message requests.
* Move correlation logic into the MessageInstanceModel class - so that logic is all in one place.
* Re-work message tests so I could wrap my simple head around what was happening - just needed an example that made sense to me.

# SpiffWorkflow Backend: Small random things
* The APIError could bomb out ugly if it hit a workflow exception with no Task Spec.
* Allow people to run commands like "flask db upgrade" without setting specific environment variables like FLASK_SESSION_SECRET_KEY everytime - they just need to add in their own /instance/config.py with their local configuration.


spiffworkflow-backend/src/spiffworkflow_backend/config/default.py
```diff
@@ -2,6 +2,8 @@
 import re
 from os import environ
 
+FLASK_SESSION_SECRET_KEY = environ.get("FLASK_SESSION_SECRET_KEY")
```
jasquat, line 5: Should we prepend with "SPIFFWORKFLOW_BACKEND_" here? If flask doesn't use it directly then it may be a good idea and looks like there my be other flask env vars we should namespace as well.

SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_definitions.py
```diff
@@ -205,16 +206,18 @@ def update_task_data(self, my_task):
...
             my_task.set_data(**payload)
 
-    def get_correlations(self, script_engine, payload):
+    def get_correlations(self, task, payload):
         correlations = {}
         for property in self.correlation_properties:
-            for key in property.correlation_keys:
-                if key not in correlations:
-                    correlations[key] = {}
-                correlations[key][property.name] = script_engine._evaluate(property.expression, payload)
```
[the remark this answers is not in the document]
danfunk: Will retain a get_correlation_keys method - and understand we may want to gather this information in this way at a future date.  However the structure is not helpful for determining if a message matches, or for storing the properties discovered through a send as properties can be shared between keys.

SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_types.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, wf_spec, name, event_definition, **kwargs):
 
     def catches(self, my_task, event_definition, correlations=None):
         if self.event_definition == event_definition:
-            return all([ correlations.get(key) == my_task.workflow.correlations.get(key) for key in correlations ])
+            return all([ str(correlations.get(key)) == str(my_task.workflow.correlations.get(key)) for key in correlations ])
```
danfunk, line 39: Still not loving this, but don't have a resultion to it at the moment.

SpiffWorkflow/SpiffWorkflow/bpmn/workflow.py
```diff
@@ -127,6 +128,10 @@ def catch(self, event_definition, correlations=None):
         should call the event_definition's reset method before executing to
         clear out a stale message.
 
+        We might be catching an event that was thrown from some other part of
+        our own workflow, and it needs to continue out, but if it originated
+        externally, we should not pass it on.
+
```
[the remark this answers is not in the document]
danfunk: not certain how to resolve this then -- could use some help identifying a failing test around this code.

SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_definitions.py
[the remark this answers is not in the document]
essweine: It is absolutely helpful -- in fact it's all you need to know outside of SpiffWorkflow.  The correlation key (name and value) is the unit of comparison to use at the interprocess level, and in the code you're proposing to replace, Spiff is doing the work of collecting the all the properties into a unit, associating it with the name, and passing both along.

This change actually doesn't make too much difference internally to SpiffWorkflow, though it's conceptually cleaner to be comparing keys rather than individual properties even internally (that's the unit of comparison).  It does make a difference in applications relying on Spiff, which is why I think it's questionable, because it makes life in spiffworkflow-backend harder.

What ought to be changed about SpiffWorkflow is what information the parser returns about messages and correlations -- if the parser returned information about the correlation keys rather than their properties, you could rely on that and let SpiffWorkflow just provide you with a name and value which you can compare.  Instead, you're removing this from SpiffWorkflow (which can handle this easily) and replacing it with more complicated code in spiffworkflow-backend.

SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_types.py
essweine: For the record, I'm completely appalled at this.

SpiffWorkflow/SpiffWorkflow/bpmn/workflow.py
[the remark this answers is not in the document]
essweine: I think I misunderstood what you were trying to accomplish.  There could be a call activity or subprocess waiting for a message generated externally and those definitely need to be passed on; however, it does make sense to check whether the message is external when deciding whether to add it back to external messages.  I don't think `event_definition.catch` needs to know about this at all, but it does make sense to prevent external messages being added back into the external message queue.

SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_definitions.py
```diff
@@ -72,9 +73,9 @@ def _throw(self, event, workflow, outer_workflow, correlations=None):
         # We also don't have a more sophisticated method for addressing events to
         # a particular process, but this at least provides a mechanism for distinguishing
         # between processes and subprocesses.
-        if self.external:
```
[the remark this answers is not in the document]
essweine: This code is pile of crap so if this change doesn't cause any issues, you might as well change it.  The new code isn't going to be any worse what's there.  Certain events are only applicable within the workflow where they originated and others aren't handled by their originating workflow and that's what this code is trying (badly) to handle.

The historical context of this is transactions (cancel events generated inside subprocesses are supposed to be sent to the outer workflow).  But sometimes I look at this code and say to myself "what the hell was I thinking?".  So no objections to changing this if necessary.

I think there could be problems at some point, but I expect they would be the result of underlying issues rather than these specific changes.

SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_types.py
burnettk: because you think the values should be exact matches even including types? you like the idea of having integer values, etc?

essweine: Are you addressing this question to me? I don't GAF about types but I do care about not silently changing data-related behavior.

If you DO care about types, then this could cause problems.

Even you're agnostic about type checking, this means that you have to be sure objects that might return `True` when equality is tested also return `True` when string equality is tested.   If you haven't implemented  a `__str__` method on your object you're going to get `obj.__repr__`, which by default is `<module.class at address>` -- probably different even for workflows internal to SpiffWorkflow and almost guaranteed to be different otherwise.

We don't implement custom data types, so we don't care, but if I was a user, I'd use them, and I'd be irritated about this.

burnettk: yeah, i was just trying to understand why you were appalled. :) i assume the reason dan changed this was because something was returning "1" and something else was returning 1 (or similar), and by his estimation they were supposed to match, so he coerced them to strings so they'd match. i'm honestly not sure if we should expect/allow correlation values to have non-string types or not. it doesn't seem terrible to me either way. i can understand why someone might be surprised by the coercing behavior. i hadn't considered the case where coercing to strings might actually cause something to NOT match that would have otherwise matched. that would be annoying. i guess you could test both to avoid that:

```` python
(val1 == val2) or (str(val1) == str(val2))
````

Comment (essweine) [reply SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_types.py:39]: Well, that solves the case where you are agnostic about types, but not the one where you do care, because you'll get a spurious match on 1 == '1'.

Even if I don't care much about type checking (I actually do care about types, I just think type checking is overrated), I'm almost certainly in the minority on this.  Since lots of people do care, I think we should do our best not to make things harder for them.

File: spiffworkflow-backend/src/spiffworkflow_backend/models/message_instance.py [hunk 6]
```diff
@@ -45,21 +46,22 @@ class MessageInstanceModel(SpiffworkflowBaseDBModel):
...
 
     id: int = db.Column(db.Integer, primary_key=True)
-    process_instance_id: int = db.Column(ForeignKey(ProcessInstanceModel.id), nullable=False)  # type: ignore
-    message_model_id: int = db.Column(ForeignKey(MessageModel.id), nullable=False)
-    message_model = relationship("MessageModel")
-    message_correlations_message_instances = relationship(
-        "MessageCorrelationMessageInstanceModel", cascade="delete"
-    )
-
+    process_instance_id: int = db.Column(ForeignKey(ProcessInstanceModel.id), nullable=True)  # type: ignore
```
jasquat, line 49: When will we have a message instance and not a process instance?

spiffworkflow-backend/src/spiffworkflow_backend/models/message_instance.py
```diff
@@ -45,21 +46,22 @@ class MessageInstanceModel(SpiffworkflowBaseDBModel):
...
+    name: str = db.Column(db.String(255))
     message_type: str = db.Column(db.String(20), nullable=False)
-    payload: str = db.Column(db.JSON)
+    payload: dict = db.Column(db.JSON)
     status: str = db.Column(db.String(20), nullable=False, default="ready")
+    user_id: int = db.Column(ForeignKey(UserModel.id), nullable=False)  # type: ignore
```
jasquat, line 54: messages are normally processed in the background so I'm not sure who the user will be unless it's the process instance initiator.

SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_definitions.py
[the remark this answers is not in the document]
danfunk: So we make this call and grab the keys instead of the properties. It sees to me that in nearly all cases we would get back an array with a single topic/converation/key.  So let's just consider that simple case for now.  The respose from this function call might look something like this: ["invoice": {"customer_name":"Sartography", "PO":12}]

From this we would create a  Message Record in the backend DB that looks roughly like this:

RECEIVE MESSAGE:
type: RECEIVE
name:  'Request Approval'
correlation_key:   ["invoice": {"customer_name":"Sartography", "PO":12}]

When A SENDING Message comes in, it will contain a message name, and a payload.  We can't create a correlation key for the send message because it the receiving Message Task or Event that defines the extraction rules that would create correlation values.

SEND MESSAGE:
type: SEND
name:  'Request Approval'
payload:    {"cust_name":"Sartography", "purchase_order":12}]

We need to match these two messages outside of SpiffWorkflow Library because we can't instantiate / deserialize all workflow instances to determine if we have a match - it is reasonable to imagine a world where there are 100s or 1000s of waiting instances.   We also need to handle API Calls that might produce messages from external systems.  So we can't depend on having any more information about a SEND message than what is shown above.

CREATING A MATCH:
We take the message payload, and use it as the context for running each of the receive messages' correlation properties extraction rules against it.  These extraction rules are specific to a Message Instance -- they are not tied to the correlation keys, or the correlation properties, they are directly associated with the receiving message task or event.

So we have to loop through the currelation _properties_, run the eval, and get a set of values.
We can then build a dictionary that we could compare with the correlation key of the recieve message to determine if we have a match.  Because we can now store the correlation key as a json structure on the message object (and not a single value in a string filed in the database), we don't have the weird issue where we are converting things to strings for comparison.  

I think this method would work, and would raise the correlation_keys up to the right level in the backend.   So I'll give this a shot now.

spiffworkflow-backend/src/spiffworkflow_backend/models/message_instance.py
danfunk: When as message is sent in over the API.

spiffworkflow-backend/src/spiffworkflow_backend/models/message_instance.py
danfunk: It is defaulting to exactly that - or in the case of an API call the user_id is set to who made that API Call -- in this way we can determine how a process instance was started.

SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_types.py
danfunk: I think I can resolve this today.

SpiffWorkflow/SpiffWorkflow/bpmn/specs/events/event_definitions.py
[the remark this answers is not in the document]
essweine: There actually doesn't need to be any calculation based on the message for catch events.  I know I said this the other day but I was wrong and should have looked at what the code I wrote actually did before making an assertion about it.

Sent messages DO contain correlation keys.  In fact, the correlation key is entirely determined by the sender (as the initiator of the conversation, they get to set the initial value).  When a message is sent, the correlation key and value is stored on the workflow.  Now this workflow can only receive messages that match the original key.

On the receiver workflow, at this point no correlations have been set, so it can accept the first matching message name.  Generally, the first message received would be a start event, so you start the process and send it the message.  The received then sets the correlation key and value on it's workflow, and at that point, it cannot accept messages that don't match the original key.

It is the message sender that does the calculations and it all happens inside Spiff.

What I'm proposing is a parse time table populated with:
process_model, message_name, start_event, throw_or_catch, correlation_key_name

You also have an instance table that looks something like this for waiting or future events to be caught:
process_instance, message_name, correlation_key_name, correlation_key_value

Without the proposed change, SpiffWorkflow sends you the message name, correlation key name, and the key value; the backend doesn't have to calculate anything.

When the backend receives a message, you can check your instance table for events that match all three returned values.

You can populate this table based on your parse time info.  When you receive a message that has no entries in your instance table, you go to the parse table and do a query for events that catch that message, and start any processes needed.  Then you insert rows for any other catch events in the original and started processes in the instance table along with the value of the correlation key sent by SpiffWorkflow.

spiffworkflow-backend/src/spiffworkflow_backend/models/message_instance.py
jasquat: But then I would think you create the process instance and then the message instance. I think it should always need a process instance.

danfunk: I changed it so that we are now recording the API Send Message in the table, not just the receive message - so that API Send Message record in the table was not sent by a process instance.  I think this is useful information to have.  That a message was sent from outside the system.

PR #144, opened 2023-02-20
backend: fix use of SSH private key for git ops

jakubgs:
Primarily this is supposed to fix the `git pull` aciton triggered by webhooks from GitHub. But in addition to that the point is to simplify that committing wrapper which has far too much in it.

Instead of passing everything as CLI arguments one should make use of already supported environment variables and the `env` argument to `subprocess` functions like `run()`. Writing extra logic in the wrapper only makes it unnecessarily complicated.
    
By passing both user, email, and the SSH options in `run_shell_command` we avoid the need to repeat the same boilerplate to provide Git config and SSH credentials.


jakubgs:
Tested on `dev.app` environemnt:

image

IT WERKS™

burnettk (approved):

PR #158, opened 2023-02-27
ci: add basic jenkinsfile for building Docker images

jakubgs:
This is the first and most basic version of a Jenkinsfile we can use.

Since we can't just hardcode things like docker name, tag, or credentials for pushing to a Docker registry it needs to be a parameter.

Here's some successful builds:

* https://ci.infra.status.im/job/spiff-workflow/job/manual-frontend/4/
* https://ci.infra.status.im/job/spiff-workflow/job/manual-backend/6/

They each built their respective component and pushed the `latest` tag for the matching Docker image.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,70 @@
+pipeline {
+  agent { label 'linux' }
+
+  options {
+    timestamps()
+    timeout(time: 20, unit: 'MINUTES')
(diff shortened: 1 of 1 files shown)


burnettk (approved):

jakubgs:
Any plans to merge this?

PR #161, opened 2023-03-01
workflow.catch() was nice, in that it is where we could send events a…

danfunk:
workflow.catch() was nice, in that it is where we could send events and messages.  With this change sending an event to catch will behave incorrectly for BPMN Messages. Only sending it to the right method will create the desired result.  It also adds a lot of additional code.   Would love a careful review of this, and any optimizations anyone can think of.


essweine:
Regarding our discussion about Spiff checking whether the key is valid: I now remember why I didn't check -- within Spiff, the key would originate with the sender, so if it couldn't be generated, there would be a script engine error when attempting to evaluate it.  I would have assumed this to also be true if an external service was calculating the key.  But I don't have issues with this and am happy to see the `external_origin` parameter removed.  I think we should merge this and move on.

PR #163, opened 2023-03-01
prefer the bpmn process name over the identifier on the logs list pag…

jasquat:
…e w/ burnettk


SpiffWorkflow/SpiffWorkflow/task.py
```diff
@@ -272,6 +272,7 @@ def log_info(self, dct=None):
         extra = dct or {}
         extra.update({
             'workflow': self.workflow.spec.name,
+            'workflow_name': self.workflow.spec.description,
```
essweine, line 275: I would suggest also renaming `workflow` to `workflow_spec`, for consistency with task spec names and descriptions.

burnettk: done

PR #164, opened 2023-03-03
Special case file data during user form submission

jbirddog:
Work on suggestion from @danfunk to move file data into its own location and put a marker in task data instead of the base64 encoded file contents. A new model was added which has a sha256 digest, blob contents and various other columns to describe the mimetype, filename, task data variable and optional list index. With the change, uploading ~50mb of user data and submitting a couple forms the task data/python environment sizes went from:

on main:

```
"python_env_size": 49736083,
"task_data_size": 99471122
```
local:
 
```
"python_env_size": 4497,
"task_data_size": 8998
```

In task data, an uploaded file now looks like:

```
"uploaded_file": "data:application/pdf;name=Free_Test_Data_10.5MB_PDF.pdf;base64,spifffiledatadigest+7a2051ffefd1eaf475dbef9fda019cb3d4a10eb8aea4c2c2a84a50a797a541bf"
```

The `data:` prefix needed to be kept so that looping back to the form would still work. Also, we need to keep base64 data at the end. A digest marker is used so we know this is not a newly uploaded file. There is a regression in the number of bytes shown in the loop back scenario that will be dealt with post merge.

The signature of `markdown_file_download_link` has changed. Now you just pass in the value of an uploaded file's task data variable:

```
uploaded_file_link = markdown_file_download_link(uploaded_file)
```

There is also a new `get_encoded_file_data` function that will transform a file data digest reference back to the fully encoded base64 string if the entire file is needed in task data - like to send through the postgres connector.

```
encoded_uploaded_file = get_encoded_file_data(uploaded_file)
```


spiffworkflow-backend/src/spiffworkflow_backend/services/process_instance_service.py
```diff
@@ -209,6 +218,97 @@ def get_users_assigned_to_task(
...
             return lane_uids
 
+    @classmethod
+    def file_data_model_for_value(
+        cls,
+        identifier: str,
+        value: str,
+        process_instance_id: int,
+    ) -> Optional[ProcessInstanceFileDataModel]:
+        if value.startswith("data:"):
```
burnettk, line 228: i wonder if this would be better as "spiffworkflowdata:". not sure how likely either are to come up in actual use.

burnettk (approved):
can't wait

spiffworkflow-backend/src/spiffworkflow_backend/services/process_instance_service.py
jbirddog: That one has to stay as "data:" since it is the `data-url` format that the frontend requires. Once its uploaded, inside the `base64,` part, we prefix our hash with `spifffiledatadigest+`.

PR #167, opened 2023-03-07
Encrypt secrets in the db

jbirddog:
Integration of [SimpleCrypt](https://pypi.org/project/Flask-Simple-Crypt/) to encrypt secrets stored in the db when they are added or updated. One ramification of this is that you cannot repeatedly hit the `Update Value` button from the UI for a secret. If you change the encrypted value to a plain text value, it will encrypt as desired. The only time a secret is decrypted (besides unit testing) is when the `secret:XYZ_SECRET` is used within service tasks. Tested with the Xero OAuth, Xero Get Currencies and Postgres Connector. Once this is merged/deployed existing secrets will need to be updated.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/routes/service_tasks_controller.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/routes/service_tasks_controller.py
@@ -43,7 +43,7 @@ def authentication_callback(
     """Authentication_callback."""
     verify_token(request.args.get("token"), force_run=True)
     response = request.args["response"]
-    SecretService().update_secret(
+    SecretService.update_secret(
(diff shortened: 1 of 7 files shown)


spiffworkflow-backend/src/spiffworkflow_backend/routes/service_tasks_controller.py
```diff
@@ -43,7 +43,7 @@ def authentication_callback(
     """Authentication_callback."""
     verify_token(request.args.get("token"), force_run=True)
     response = request.args["response"]
-    SecretService().update_secret(
+    SecretService.update_secret(
```
burnettk, line 46: thanks

burnettk (approved):

PR #180, opened 2023-03-14
Safe asserts

jbirddog:
The implementation of this is not as beautiful as it would be in languages with macros, but still achieves the end goal of allowing assertions in code that pass `./bin/run_pyl` that hit during local development but result in logging/sentry notifications in other environments. This would allow for extra debugging information of assumptions that don't hold in non dev environments - especially when trying to debug issues in the background worker, etc.

The assertion error is reported at the call site and would occur event if assertions are disabled. Example [link](https://sartography.sentry.io/issues/4005529436/?environment=local_development&project=4503966086529024&query=AssertionError&referrer=issue-stream&statsPeriod=14d).

Logs show:

```
AssertionError: The current thread has not obtained a lock for this process instance (54).
Traceback (most recent call last):
  File "/app/src/spiffworkflow_backend/routes/process_instances_controller.py", line 131, in process_instance_run
    processor.do_engine_steps(save=True)
  File "/app/src/spiffworkflow_backend/services/process_instance_processor.py", line 1649, in do_engine_steps
    execution_service.do_engine_steps(exit_at, save)
  File "/app/src/spiffworkflow_backend/services/workflow_execution_service.py", line 297, in do_engine_steps
    raise AssertionError(
AssertionError: The current thread has not obtained a lock for this process instance (54).


```

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/services/assertion_service.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/services/assertion_service.py
@@ -1,0 +1,16 @@
+"""Assertion_service."""
+
+import contextlib
+import sentry_sdk
+from flask import current_app
+
+@contextlib.contextmanager
+def safe_assertion(condition: bool) -> None:
+    try:
+        yield
+    except AssertionError as e:
+        if not condition:
+            sentry_sdk.capture_exception(e)
+            current_app.logger.exception(e)
+            if current_app.config["ENV_IDENTIFIER"] == "local_development":
+                raise e
--- a/spiffworkflow-backend/src/spiffworkflow_backend/services/assertion_service.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/services/assertion_service.py
@@ -1,13 +1,15 @@
 """Assertion_service."""
-
 import contextlib
+from typing import Generator
+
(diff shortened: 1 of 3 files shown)


spiffworkflow-backend/src/spiffworkflow_backend/services/assertion_service.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@contextlib.contextmanager
+def safe_assertion(condition: bool) -> Generator[bool, None, None]:
+    try:
+        yield True
+    except AssertionError as e:
+        if not condition:
+            sentry_sdk.capture_exception(e)
+            current_app.logger.exception(e)
```
burnettk, line 16: i'm not very confident in this, but sentry seems to override logger.exception in some cases and send the errors to sentry. just pointing this out in case we see doubled errors.

burnettk (approved):

spiffworkflow-backend/src/spiffworkflow_backend/services/assertion_service.py
jbirddog: Funny thing I saw that when logging the exception as error, which I would have thought would have been ok. But yes will keep an eye on it and switch to info if needed.

PR #186, opened 2023-03-17
UAT bug fix: suspended/terminated instances automagically resume

jbirddog:
The root cause here was a lock was not obtained in the process_instance_queue before suspending/terminating/resuming an instance manually, Because of this the queue was out of sync and would run the process instance which resulted in its status being updated. Fix is to take a lock when performing these actions. Also added a safe_assert to make sure the assumption holds that when the bg processor runs the process instance status is what is expects it to be.

Also moved suspend/resume to instance methods for locking parity.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/services/process_instance_processor.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/services/process_instance_processor.py
@@ -1249,7 +1249,7 @@ class ProcessInstanceProcessor:
         self.add_step()
         self.save()
         # Saving the workflow seems to reset the status
-        self.suspend(self.process_instance_model)
+        self.suspend()
 
     def reset_process(self, spiff_step: int) -> None:
         """Reset a process to an earlier state."""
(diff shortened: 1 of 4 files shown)


spiffworkflow-backend/src/spiffworkflow_backend/services/process_instance_processor.py
```diff
@@ -1900,16 +1900,14 @@ def terminate(self) -> None:
         db.session.add(self.process_instance_model)
         db.session.commit()
 
-    @classmethod
-    def suspend(cls, process_instance: ProcessInstanceModel) -> None:
+    def suspend(self) -> None:
```
jasquat, line 1903: Any reason to not keep these as classmethods and change the lock method to a classmethod. It doesn''t seem like the lock method needs anything from the processor except for the process instance anymore which can be passed in.

jbirddog: Good point - sorry I missed this before hitting merge. Most of the callers (granted most are in tests) already have the instance, but yes. Really want I want to move to in places like this are a `with locked_*():` type of call so we don't necessarily need any instance and also don't need all the try/catch/except stuff. Will put that down on the todos.

PR #192, opened 2023-03-23
ci: add discord notifications to at build success

jakubgs:
Still needs testing.


jakubgs:
Appears to work:

image

But will need some extra setup on CI side once merged.

burnettk (approved):

PR #195, opened 2023-03-27
Handle the multiple single file upload widget case

jbirddog:
In the case of the "multiple single file upload" widget, instead of being an array of data urls, it was an array of dictionaries with a single key and a data url. This caused the file content swapping not to occur. Fix is to detect this file upload style and make it appear as if it was a single multiple file upload (array of data urls). This takes the file contents out of task data as expected and simplifies the downstream code - especially the file linking.


jbirddog:
Drafting - realized this breaks looping back to the form.

PR #199, opened 2023-03-31
Trip safe asserts in tests, various process instance queue improvements

jbirddog:
Driven by the need to have the safe asserts trip when unit tests are running - which would have allowed us to see the message start event error where the created process instance was not added to the queue. The unit test environment was added to to the check to raise the `AssertionError`s in local development.

While in there I ran across the possibility of an exception during user form submission causing the process instance to remain locked. Combined with the tribal knowledge of how the enqueue/lock/unlock dance is done - decided to simplify this a bit for the callers.

A new `enqueue_new_process_instance` method was added that is called automatically any time a process instance model is created through the sanctioned method(s). The existing `enqueue` and `dequeue` methods have been prefixed with an underscore to let callers know they probably shouldn't be calling these directly. A new context manager is added that will handle exception handling, reentering and balancing of the locks. The process instance processor's lock/unlock methods have been removed in favor of the context manager.

If you are calling do_engine_steps the locking is handled for you. If you are doing something else you will need to be inside the context manager.

Also removed the retry/sleep logic since we no longer lock all user forms to check for timers.


burnettk (approved):
awesome

PR #218, opened 2023-04-19
Called elements

jbirddog:
~~Requires [SpiffWorkflow PR](https://github.com/sartography/SpiffWorkflow/pull/316) and poetry update before merging.~~

Adds support for caching/querying processes which call another process. The exposed api returns the same structure as `process_list`.


burnettk (approved):

spiffworkflow-backend/src/spiffworkflow_backend/api.yml
```diff
@@ -519,6 +519,31 @@ paths:
...
+  /processes/{bpmn_process_identifier}/callers:
+    parameters:
+      - name: bpmn_process_identifier
+        in: path
+        required: true
+        description: the modified process model id
+        schema:
+          type: string
+    get:
+      operationId: spiffworkflow_backend.routes.process_api_blueprint.process_callers
```
burnettk, line 531: process_caller_list might be slightly more conventional for the function name.

spiffworkflow-backend/src/spiffworkflow_backend/models/process_caller.py
```diff
@@ -0,0 +1,12 @@
+"""ProcessCaller_model."""
+from spiffworkflow_backend.models.db import db
+from spiffworkflow_backend.models.db import SpiffworkflowBaseDBModel
+
+
+class ProcessCallerCache(SpiffworkflowBaseDBModel):
```
burnettk, line 6: we almost always but not always use `Model` as the suffix for db models.

spiffworkflow-backend/src/spiffworkflow_backend/routes/process_api_blueprint.py
```diff
@@ -77,6 +78,14 @@ def process_list() -> Any:
     return SpecReferenceSchema(many=True).dump(references)
 
 
+def process_callers(bpmn_process_identifier: str) -> Any:
+    callers = ProcessCallerService.callers(bpmn_process_identifier)
+    references = (
+        SpecReferenceCache.query.filter_by(type="process").filter(SpecReferenceCache.identifier.in_(callers)).all()
+    )
+    return SpecReferenceSchema(many=True).dump(references)
```
burnettk, line 86: we're trying to move away from marshmallow schemas in favor of dataclasses, etc, for serialization. no action required here, since i know SpecReferenceCache already had the Schema, but just for new models for the future.

burnettk (approved):

PR #223, opened 2023-04-22
* Scroll to the top on each update of a process result.

danfunk:
* When displaying an error, scroll to the top of the page.
* Rename a few of the filter titles to better match the titles of the table on the process instances list page.
* Add error_message to the task object - so if a task is in an error state and has a message, we send it to the front end.
* Handle errors on the Interstitial page - so if an error happens during execution it is diplayed, and if an error happened previously, we show it.
* Return the last Errored Tasks if no READY or WAITING tasks exists when calling the Interstitial endpoint.


spiffworkflow-frontend/src/components/ProcessModelSearch.tsx
```diff
@@ -65,7 +65,7 @@ export default function ProcessModelSearch({
         return null;
       }}
       shouldFilterItem={shouldFilterProcessModel}
-      placeholder="Choose a process model"
```
burnettk, line 68: This one is kind of heartbreaking

burnettk (approved):

PR #230, opened 2023-04-27
Support connectors that return logs

jbirddog:
Initially this is for the http get/post v2 connectors which return a `spiff__logs` key along with their response. If this format is detected we merge in the logs with the backend logs and promote the nested `api_response` as the response. This keeps logs out of task data and also allows for temporarily switching to the v2 format if you just want to see logs while debugging. Marking as draft until after the status demo.


burnettk (approved):
slick, so we can seamlessly start using v2

PR #238, opened 2023-05-03
boundary event suspend manual complete

danfunk:
* Test and updates to assure that when a task has a boundary event, and you return to that event, and then progress one step, you don't get stuck with a task that can't ever be completed.
* Let SpiffWorkflow determine what tasks we need to update in the DB using the task_state_change date on the tasks.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/services/process_instance_processor.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/services/process_instance_processor.py
@@ -1115,59 +1115,54 @@ class ProcessInstanceProcessor:
 
     def manual_complete_task(self, task_id: str, execute: bool) -> None:
         """Mark the task complete optionally executing it."""
-        spiff_tasks_updated = {}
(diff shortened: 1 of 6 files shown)


spiffworkflow-backend/src/spiffworkflow_backend/services/process_instance_processor.py
```diff
@@ -1320,6 +1323,11 @@ def reset_process(cls, process_instance: ProcessInstanceModel, to_task_guid: str
         db.session.commit()
 
         processor = ProcessInstanceProcessor(process_instance)
+
+        # If this as a boundary event parent, run it, so we get back to an active task.
+        if is_boundary_parent:
+            processor.do_engine_steps(execution_strategy_name="one_at_a_time")
```
burnettk, line 1329: nice use of the one at a time strategy

burnettk (approved):

jasquat (approved):

PR #243, opened 2023-05-05
Feature/interstitial do not update pi status

jasquat:
This avoids changing the process instance status when going to the interstitial page when the status is suspended. It also disallows the page from running error statuses as well. Before the interstitial page would run an instance in either state and update that state.


spiffworkflow-backend/src/spiffworkflow_backend/routes/tasks_controller.py
```diff
@@ -400,6 +401,11 @@ def render_instructions(spiff_task: SpiffTask) -> str:
         extensions = TaskService.get_extensions_from_task_model(task_model)
         return _render_instructions_for_end_user(task_model, extensions)
 
+    def render_data(return_type: str, entity: Union[ApiError, Task]) -> str:
```
danfunk, line 404: nice introduction of return type here.  Much cleaner

danfunk (approved):

burnettk (approved):

PR #253, opened 2023-05-10
Type ahead cleanup

jbirddog:
Cleanup around the "typeahead" for pp2 prep. Renamed the url/env vars to take out the hypen/underscore, removed the title from the combobox and stopped error suppression since upon further testing it didn't really work. For now we will get a popup if the typeahead server is down. Issue is if there is no selected item, the field has no value, so what you think is working isn't really. So better to stop it early and let us get a screenshot of the error.


jbirddog:
[Companion pr](https://github.com/jbirddog/typeahead/pull/6)

PR #264, opened 2023-05-17
Force all pages back to the top, rather than just a few of them.

danfunk:


danfunk:
Look Maw!  No console logs!

burnettk (approved):

PR #273, opened 2023-05-23
What is causing this error

jbirddog:
Initializing the processor here causes some interesting errors. This is what was causing the issues with the timer start events.


jbirddog:
Looks like the way some tests are configured, they are not compatible with being created via the processor.

PR #279, opened 2023-05-25
Editor compose

jbirddog:
Adds the start of the "editor" via a `editor.docker-compose.yml` and a set of scripts to run/stop and update the editor images. Once auth-less/editor parameters are available we can set them in this docker-compose file to see how the editor looks. Once merged you can:

`./bin/run_editor ~/some/bpmn/dir`


danfunk (approved):
seems reasonable

PR #285, opened 2023-05-30
Cycle Timer Start Event Support

jbirddog:
Adds support for cycle timer start events. Once the first instance completes, if there are any remaining cycles a new instance is scheduled after the cycle duration. The process initiator from the original instance is used for the cycle instances to prevent some privilege escalation. If there were a properly privileged system user that the background processor we could look at using that instead.


spiffworkflow-backend/src/spiffworkflow_backend/models/process_model_cycle.py
```diff
@@ -0,0 +1,20 @@
+"""Process_instance_queue."""
```
burnettk, line 1: https://github.com/sartography/spiff-arena/commit/274892125d7b8211bb7024696885767409dc6a59

burnettk (approved):

spiffworkflow-backend/src/spiffworkflow_backend/models/process_model_cycle.py
jbirddog: nice.

PR #291, opened 2023-05-31
allow not redirecting to home page at all in interstitial component a…

jasquat:
…nd do not error out if next task does not exist

### DIFF
--- a/spiffworkflow-frontend/src/components/ProcessInterstitial.tsx
+++ b/spiffworkflow-frontend/src/components/ProcessInterstitial.tsx
@@ -15,12 +15,14 @@ type OwnProps = {
   processInstanceId: number;
   processInstanceShowPageUrl: string;
   allowRedirect: boolean;
+  redirectToHomeWhenNoTasks?: boolean;
 };
 
 export default function ProcessInterstitial({
   processInstanceId,
   allowRedirect,
   processInstanceShowPageUrl,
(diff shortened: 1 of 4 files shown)


burnettk (approved):

spiffworkflow-frontend/src/components/ProcessInstanceListTable.tsx
```diff
@@ -490,6 +490,7 @@ export default function ProcessInstanceListTable({
         HttpService.makeCallToBackend({
           path: `/process-instances/report-metadata${queryParamString}`,
           successCallback: getProcessInstances,
+          onUnauthorized: stopRefreshing,
```
danfunk, line 493: no more recursive popups!

spiffworkflow-frontend/src/components/ProcessInterstitial.tsx
```diff
@@ -15,12 +15,14 @@ type OwnProps = {
   processInstanceId: number;
   processInstanceShowPageUrl: string;
   allowRedirect: boolean;
```
danfunk, line 17: why not just use allowRedirect?

jasquat: Oh we could use that instead. I thought that was for redirecting to process instance page but i guess it could be used for both.

danfunk (approved):

PR #335, opened 2023-06-15
parallel tests with xdist

burnettk:
`./bin/tests-par` runs about four times faster than `poet test` for me.


spiffworkflow-backend/bin/tests-par
```diff
@@ -0,0 +1,21 @@
...
+set -o errtrace -o errexit -o nounset -o pipefail
+
+if [[ ! -f ./src/instance/db_unit_testing_gw0.sqlite3 ]] ; then
+  >&2 echo -e "ERROR: please run the following command first in order to set up and migrate the sqlite unit_testing database:\n\n\tSPIFFWORKFLOW_BACKEND_DATABASE_TYPE=sqlite ./bin/recreate_db clean"
+  exit 1
+fi
+
+# check if python package hey is installed
+if ! python -c "import xdist" &>/dev/null; then
+  >&2 echo -e "ERROR: please install the python package pytest-xdist first:\n\n\tpip install pytest-xdist"
```
jbirddog, line 17: Could this be a dev dependency via poetry?

burnettk:
yeah, true, and that shouldn't affect prod images.

spiffworkflow-backend/bin/tests-par
```diff
@@ -0,0 +1,21 @@
...
+
+function error_handler() {
+  >&2 echo "Exited with BAD EXIT CODE '${2}' in ${0} script at line: ${1}."
+  exit "$2"
+}
+trap 'error_handler ${LINENO} $?' ERR
+set -o errtrace -o errexit -o nounset -o pipefail
+
+if [[ ! -f ./src/instance/db_unit_testing_gw0.sqlite3 ]] ; then
+  >&2 echo -e "ERROR: please run the following command first in order to set up and migrate the sqlite unit_testing database:\n\n\tSPIFFWORKFLOW_BACKEND_DATABASE_TYPE=sqlite ./bin/recreate_db clean"
```
jbirddog, line 11: Also any reason not to just run this command in this scenario? btw very pumped about this landing.

jbirddog (approved):

spiffworkflow-backend/bin/tests-par
burnettk: i think it'll clear the development database, too, so might be a bit dangerous. could prompt, i guess.

jbirddog: Gotcha - if maybe dangerous then sure either way is fine. Less code not to prompt I suppose.

PR #337, opened 2023-06-15
Feature/interstitial summary

danfunk:
* Process instance logs and messages are now components rather than pages, and are included within tabs on the process instance page, along with the diagram.
* Removed the Zoom and Move modules when showing the readonly Diagram.  Assured this readonly view is resized to fit the space when possible.
* Checkbox Widget no longer displays a duplicate label.
* CSS Tweaks
  * All pages are limited to a max display width of 1440, with auto margins to center the main content on the page.
  * "Show" pages, like ProcessInstanceShow, TaskShow have the primary content limited to 1000, also with auto-margins.
  * Paragraphs, headings, blockquotes, list items are limited to a width of 640.
  * Reduced margin bottom on all breadcrumbs.
  * Slightly reduced the width and margin of tiles
  * ordered lists and unordered lists show numbers and bullets now.
* End user Instructions component can, optionally, auto-collapse, so that only a portion is displayed, along with a toggle.  This is how it is set up for the ProcessInstanceShow page.
* Greatly reduced the lag in the interstitial page when doing a re-direct.

### DIFF
--- a/spiffworkflow-frontend/src/routes/AdminRoutes.tsx
+++ b/spiffworkflow-frontend/src/routes/AdminRoutes.tsx
@@ -16,8 +16,6 @@ import ProcessInstanceReportList from './ProcessInstanceReportList';
 import ProcessInstanceReportNew from './ProcessInstanceReportNew';
 import ProcessInstanceReportEdit from './ProcessInstanceReportEdit';
 import ReactFormEditor from './ReactFormEditor';
-import ProcessInstanceLogList from '../components/ProcessInstanceLogList';
-import MessageInstanceList from '../components/MessageInstanceList';
 import Configuration from './Configuration';
 import JsonSchemaFormBuilder from './JsonSchemaFormBuilder';
 import ProcessModelNewExperimental from './ProcessModelNewExperimental';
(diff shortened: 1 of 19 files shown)


spiffworkflow-frontend/src/routes/ProcessInstanceShow.tsx
```diff
@@ -1099,71 +1028,114 @@ export default function ProcessInstanceShow({ variant }: OwnProps) {
       params.process_model_id ? params.process_model_id : ''
     );
 
+    const getTabs = () => {
```
burnettk, line 1031: any time you call `ability.can`, it can suffer from the problem that the permissions are not loaded yet and then it says you don't have access but you actually do, so it's safest to wrap the call in a check to ensure that permissionsLoaded is true. i have added one to line 1026 in ProcessInstanceShow and it didn't obviously break anything. this check is not required when using the jsx `<Can>` component for some reason, which is why we have a bias for that construct, but it's not always possible and it can definitely be a lot more natural to avoid it.

burnettk (approved):

spiffworkflow-frontend/src/routes/AdminRoutes.tsx
```diff
@@ -132,7 +122,6 @@ export default function AdminRoutes() {
           path="process-instances/all"
           element={<ProcessInstanceList variant="all" />}
         />
-        <Route path="messages" element={<MessageInstanceList />} />
```
danfunk, line 135: Need to add this back in.

spiffworkflow-frontend/src/routes/ProcessInstanceShow.tsx
danfunk: AH - that makes a ton of sense - thank you for the explanation.

PR #338, opened 2023-06-15
throw an error if backend gives a 500 from interstitial page w/ burnettk

jasquat:


jasquat:
Closing in favor of https://github.com/sartography/spiff-arena/pull/340 since this was merged into it.

PR #339, opened 2023-06-16
when loading task data from the server in task show, convert null bac…

jasquat:
…k to undefined so rjsf does not validate the values w/ burnettk


jasquat:
Closing in favor of https://github.com/sartography/spiff-arena/pull/342

PR #350, opened 2023-06-25
Spiffdemo getting started section

usama9500:
Getting Started
Using SpiffDemo.org 
Logging in 
Minimal Example


Phillana26 (approved):
- May just try and zoom in on the process when you take a screen shot. I think the writing is quite small and the focus of the image should be the process model and zoomed in so its easier to read. 
- Mention that the name can be left empty but best practice is to give it a name that described the process and can be identified. 
- clearly state which fields are optional/mandatory
-links to sections explaining the BPMN element in the documentation will be helpful

PR #400, opened 2023-07-19
Basic scaffolding for onboarding

jbirddog:
Adds a shim view to the frontend to allow changing the homepage, to a degree based on results from an onboarding process model that is called, if it exists. If the onboarding model has human tasks they will be handled normally. To support running the onboarding process `process_instance_create` and `process_instance_run` were refactored and a `process_instance_start` was added that does the two steps in one.

Likely may go back and add message support later, can either be the first check or done after the known location check that current exists fails.


spiffworkflow-frontend/src/routes/OnboardingView.tsx
```diff
@@ -0,0 +1,42 @@
...
+      if (onboarding.type === 'default_view') {
+        if (onboarding.value === 'my_tasks') {
+          return <MyTasks />;
+        }
+      } else if (
+        onboarding.type === 'user_input_required' &&
+        onboarding.process_instance_id &&
+        onboarding.task_id
+      ) {
+        console.log('HERE');
```
burnettk, line 31: console straggler

burnettk (approved):
awesome

spiffworkflow-frontend/src/routes/OnboardingView.tsx
jbirddog: good catch

spiffworkflow-backend/src/spiffworkflow_backend/routes/process_instances_controller.py
```diff
@@ -134,6 +149,14 @@ def process_instance_run(
     return make_response(jsonify(process_instance), 200)
 
 
+def process_instance_start(
```
jasquat, line 152: Should we underscore this? That's what we've been doing when a method is called by a controller and not by the api. Helps to show what are actually api methods and what aren't.

jasquat (approved):
How hard would it be to add an integration test for this? Also, do we need a permission set for the "/onboarding" endpoint? Is it apart of BASIC permissions?

spiffworkflow-backend/src/spiffworkflow_backend/routes/process_instances_controller.py
jbirddog: No problems with that.

jbirddog:
> How hard would it be to add an integration test for this? Also, do we need a permission set for the "/onboarding" endpoint? Is it apart of BASIC permissions?

I'm not sure honestly, i think it would be checking if the homepage did or did not do something different - maybe a cypress test? Don't have a confident answer but not opposed to a test either.

jbirddog:
> Is it apart of BASIC permissions?

Added read for /onboarding to the basic permissions

jbirddog:
Adding an integration test for when an onboarding model is present and when one isn't.

PR #403, opened 2023-07-19
make perspective dropdown bigger

burnettk:
make perspective selection dropdown component only truncate text when it actually doesn't fit.


jasquat:
FWIW the report identifier column can only be 50 characters atm so we don't actually need the field that large or we should actually increase the column restriction to 255.

jasquat (approved):

PR #417, opened 2023-07-25
Don't schedule a cycle with 0 second duration

jbirddog:
If a cycle has a duration of 0 seconds don't schedule it else the system will continuously run the process. If the process equivalent of a while true loop is really desired, configure the cycle timer to fire every second. This should help with the cycle timer issue (5ce8a7a921d14e61bdb75d7907c52bbb) but is not a true root cause fix.


jasquat (approved):

jasquat:
Should we raise as well? Or is 0 expected and we should just ignore it? We may not even know the answer unless we raise.

jbirddog:
> Should we raise as well? Or is 0 expected and we should just ignore it? We may not even know the answer unless we raise.

Sure raise sounds fine to me. Updated.

PR #427, opened 2023-08-02
updated bpmn-js-spiffworkflow for mi payload bugfix w/ burnettk

jasquat:
See https://github.com/sartography/bpmn-js-spiffworkflow/pull/37


jasquat:
This was actually committed with https://github.com/sartography/spiff-arena/commit/875db6b1b9d92234c5e36f1ac7e4d82641a587a4.

PR #430, opened 2023-08-03
Minor tweaks t othe onboarding page. 

danfunk:
* Display the instructions_for_user directly on the home page (no redirects to alternate pages)
* Moved the onboarding view above the tabs on the home page.
* Added a "times_executed_by_user" script - not sure if we should keep it, but maybe handy.
* Assure that the display_name for the user is returned consistently (it was not being returned by the get_current_user function).

With @jbirddog

Next up will look at getting this to work using messages rather than with a known path, but wanted to get a good message out there.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/models/user.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/models/user.py
@@ -27,12 +27,12 @@ class UserModel(SpiffworkflowBaseDBModel):
 
     id: int = db.Column(db.Integer, primary_key=True)
     username: str = db.Column(db.String(255), nullable=False, unique=True)
-    email = db.Column(db.String(255), index=True)
+    email: str = db.Column(db.String(255), index=True)
 
     service = db.Column(db.String(255), nullable=False, unique=False, index=True)  # not 'openid' -- google, aws
     service_id = db.Column(db.String(255), nullable=False, unique=False, index=True)
 
-    display_name = db.Column(db.String(255))
+    display_name: str = db.Column(db.String(255))
     tenant_specific_field_1: str | None = db.Column(db.String(255))
     tenant_specific_field_2: str | None = db.Column(db.String(255))
     tenant_specific_field_3: str | None = db.Column(db.String(255))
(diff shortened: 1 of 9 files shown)


spiffworkflow-backend/src/spiffworkflow_backend/models/user.py
```diff
@@ -27,12 +27,12 @@ class UserModel(SpiffworkflowBaseDBModel):
 
     id: int = db.Column(db.Integer, primary_key=True)
     username: str = db.Column(db.String(255), nullable=False, unique=True)
-    email = db.Column(db.String(255), index=True)
+    email: str = db.Column(db.String(255), index=True)
```
danfunk, line 30: Adding this type hint causes the email and display_name to come back in the json serialization of g.user --

spiffworkflow-frontend/src/routes/OnboardingView.tsx
```diff
@@ -18,11 +19,23 @@ export default function OnboardingView() {
...
-    if (onboarding) {
+    if (onboarding && onboarding.instructions.length > 0) {
+      return (
+        <MDEditor.Markdown
+          className="onboarding"
+          linkTarget="_blank"
+          source={onboarding.instructions}
+        />
+      );
+      /*
```
danfunk, line 30: Leaving this in for now - as I suspect we want to navigate to the workflow if it is on a user task and is not complete - but need to think more about how to handle this navigation ...

spiffworkflow-frontend/src/routes/OnboardingView.tsx
```diff
@@ -18,11 +19,23 @@ export default function OnboardingView() {
...
       if (onboarding.type === 'default_view') {
         if (onboarding.value === 'my_tasks') {
           return <MyTasks />;
         }
+      } else if (
+          onboarding.type === 'user_input_required'
+      ) {
+        console.log("onboarding");
```
burnettk, line 38: console

burnettk (approved):

jasquat (approved):

PR #432, opened 2023-08-03
Support OAuth without Connector Proxy

jbirddog:
Adds support to do the oauth dance without needing to go through a connector proxy. The oauth params are currently hardcoded into a dictionary with github as a working example. Secrets are used to store the client id/secret. Once those secrets are added you can oauth against github and make a call to get user information. Other calls would work if the scope was expanded. Besides a small refactor to get secrets, the existing connector proxy flow is untouched. To test the github oauth, set up a oauth app, add the secrets, then you get use the http get connector with:

Headers:

`{ "Authorization": "Bearer SPIFF_SECRET:github_oauth" }`

URL:

`"https://api.github.com/user"`


jasquat:
This is included in https://github.com/sartography/spiff-arena/pull/434.

PR #434, opened 2023-08-08
Oauth admin

jbirddog:
This builds on #432, which should be merged before this. Adds a local configuration json editor under `Configuration -> Authentications` which allows the admin to describe OAuth configurations. For example the github config is:

```
{
    "github": {
        "consumer_key": "SPIFF_SECRET:GITHUB_CONSUMER_KEY",
        "consumer_secret": "SPIFF_SECRET:GITHUB_CONSUMER_SECRET",
        "request_token_params": {
            "scope": "user:email"
        },
        "base_url": "https://api.github.com/",
        "request_token_url": null,
        "access_token_method": "POST",
        "access_token_url": "https://github.com/login/oauth/access_token",
        "authorize_url": "https://github.com/login/oauth/authorize"
    }
}
```

These will be added to the table of available authentications. Also added a column for the source of the authentication (connector proxy or local config) so it wasn't possibly confusing as to why some options were showing up that were not in the local config.

Screenshot from 2023-08-08 13-58-56


burnettk:
@jbirddog do you happen to remember where SPIFF_SECRET:GITHUB_CONSUMER_KEY and GITHUB_CONSUMER_SECRET come from before you add them as secrets? i'm observing that on the test.app environment, it's using something like your sample config from here, but those secrets don't exist.

jbirddog:
I went through and set them up the in the github oauth settings if I recall. I have them locally as well, can provide in another channel.

PR #439, opened 2023-08-11
Threads

danfunk:
The default behavior for all execution strategies is, IF there is more than one ready engine task, run all engine tasks in parallel. This takes advantage of the GIL - so we are not at all worried about shared memory issues.  So this won't help with processor intensive issues, but it WILL help with many simultaneous IO bound issues - so you can gain performance improvements by executing many services tasks in parallel, so you aren't waiting for one to return before calling the next - you can make all the calls simultaneously and wait for the results.  

We also made sure that the Task table is cleaned up on save - so that when you save the process instance, any pruned tasks in the spiff workflow library model are also pruned from the database.

with @jbirddog, @jlantz, @burnettk


jbirddog (approved):
Not a big deal but just thinking, believe we can drop the `greedy` strategy now as well and just use `run_current_ready_tasks` in its place.

PR #447, opened 2023-08-21
Feature/guest form submission

jasquat:
This adds in support to complete specified human tasks with a guest user.

To test:
1. Create a process model with a manual or user task and click the allow guest checkbox
1. Start the process model with the same user and go to the process instance show page to get the guid of the human task
1. Use that guid to create a url like: "[domain]/tasks/[process_instance_id]/[task_guid]" and navigate there from a browser that is not logged into spiff
1. This user should be able to complete the task and will be given the next allowed guest task if there is one
1. If the guest user tries to navigate anywhere that is not the task show page, they should be taken to the login screen


spiffworkflow-backend/src/spiffworkflow_backend/models/task.py
```diff
@@ -56,6 +56,8 @@ class TaskModel(SpiffworkflowBaseDBModel):
     human_tasks = relationship("HumanTaskModel", back_populates="task_model", cascade="delete")
     process_instance_id: int = db.Column(ForeignKey("process_instance.id"), nullable=False, index=True)
 
+    allow_guest: bool = db.Column(db.Boolean())
```
burnettk, line 59: could consider killing this cache column if the query filter isn't truly necessary.

spiffworkflow-backend/src/spiffworkflow_backend/services/user_service.py
```diff
@@ -82,6 +83,14 @@ def current_user() -> Any:
             raise ApiError("logged_out", "You are no longer logged in.", status_code=401)
         return g.user
 
+    @classmethod
+    def is_logged_in_as_guest_user(cls) -> bool:
```
burnettk, line 87: i think we can kill this unless it's being used by something meta-programmed.

burnettk (approved):
it seems this one also needs the magic to combine the db migrations.

spiffworkflow-backend/src/spiffworkflow_backend/services/user_service.py
jasquat: Looks like not. I think was using it but seems no longer necessary.

spiffworkflow-backend/src/spiffworkflow_backend/models/task.py
jasquat: This has been removed.

PR #490, opened 2023-09-14
File based generic json data store

jbirddog:
Adds another flavor of the generic json data store, this uses a file on disk as persistence instead of a db table. This allows a json file to be included along with a process model. At this time the file has to be in the location of the executing process model. Soon this requirement will be lifted when support for finding files in higher/shared directories is added.


jasquat (approved):
Looks fine. Should a unit test be added for this? Or do old tests handle that?

jbirddog:
> Looks fine. Should a unit test be added for this? Or do old tests handle that?

Don't mind looking into adding a test for it.

PR #516, opened 2023-09-28
Refactor finding process models to support finding other file types 

jbirddog:
Refactors the way process models are located so that other file types, such as data stores can be located as well. This allows data stores to be populated in the reference cache for later "up searching".


spiffworkflow-backend/src/spiffworkflow_backend/services/process_model_service.py
```diff
@@ -180,13 +184,18 @@ def get_process_models(
             awesome_id = process_group_id.replace("/", os.sep)
             root_path = os.path.join(root_path, awesome_id)
 
-        process_model_glob = os.path.join(root_path, "*", "process_model.json")
```
burnettk, line 183: do you happen to know whether this may be faster or slower?

burnettk (approved):

spiffworkflow-backend/src/spiffworkflow_backend/services/process_model_service.py
jbirddog: I read that in python 3.5+ os.walk became faster due to it switching over to os.scandir which reduced the number of stat calls. Just tried porting an old quick and dirty benchmark from stack overflow and got (running on sample process models):

```
>>> glob_time()
glob.glob: 0.0171s, 613 files found
>>> walk_time()
os.walk: 0.0156s, 1339 files found
```

The os.walk is negligibly faster but also found more files, so if my glob was 1:1 i'm not sure what the time would be but I assume slower.  Main benefit though is we can find process models and whatever other files we want to index in one read of the file system vs multiple globs or one gnarly glob.

Old benchmark results from SO:

```
os.walk: 3.6592s, 1326787 files found
glob.glob: 2.0133s, 1326786 files found
```

https://stackoverflow.com/questions/8931099/quicker-to-os-walk-or-glob

PR #520, opened 2023-09-29
Upsearch for data stores

jbirddog:
Converts json and json file data stores to "upsearch" for an entry using the new reference cache. This will allow sharing json data stores between process models contained in a similar hierarchy. Currently data stores cannot be added to groups in the UI, so that will need to be added for this to work without manually placing the file on the file system. The same "upsearch" approach can be used for anything in the reference cache - process models and dmn files for instance, but is only integrated for json data stores in this pr.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/data_stores/json.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/data_stores/json.py
@@ -24,6 +24,7 @@ def _data_store_filename(name: str) -> str:
 def _data_store_exists_at_location(location: str, name: str) -> bool:
     return FileSystemService.file_exists_at_relative_path(location, _data_store_filename(name))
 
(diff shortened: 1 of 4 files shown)


burnettk (approved):

madhurrya:
@jbirddog Is there anything we can test in the UI related to this PR?

dinithihj:
@jbirddog I'm trying to test this and have some confusion. 

1. If I understand correctly with this feature, the data stores created at the upper level (process group level?) can be accessed by the process models created within that process group. Please correct me if I'm wrong.
2. In the first comment you mentioned that "Currently data stores cannot be added to groups in the UI". So is there anyway I can test this by adding data stores to groups with or without using UI?
3. If the answer is No to 2nd question, is there anything I can test related to this ticket?

PR #530, opened 2023-10-06
Feature/regex support in permissions

jasquat:
This adds support in permissions to set the username to a regex so it can be used to add users matching that regex to a group.

Includes:
* regex support in permissions - prepend username with "REGEX:" to use
* removes default "*"  support from UserGroupAssignmentWaitingModel since it duplicates the everybody group idea and now we can use regexes
* some additional refactoring to clean up the auth / login flow
  * changes name of routes/user.py to routes/authentication_controller.py

To test;
* in the permission process model or in the yaml file, create a group assignment where the username is something like: `REGEX:.*@example.com`
  * group name doesn't matter
* give that group some permission easy to check
* run the set permissions process model
* log in with a user that matches the regex and ensure they get the permission
* delete this group assignment from the process model or yaml file
* run the set permissions process model
* log in with the user again and they should no longer have that permission


burnettk (approved):

madhurrya:
Tested using this.
image
image

PR #534, opened 2023-10-10
Feature/metadata column display type

jasquat:
This adds in the ability to define how metadata should display when creating a report. Right now only "Date / time" and "Duration" are supported.

This adds in the ability to turn off "with-tasks-for-me" from a report on the All instances list to make it easier for a user to share reports with someone who has access to the All screen.

### DIFF
--- a/spiffworkflow-frontend/src/components/ProcessInstanceListTable.tsx
+++ b/spiffworkflow-frontend/src/components/ProcessInstanceListTable.tsx
@@ -71,6 +71,7 @@ import {
   ErrorForDisplay,
   PermissionsToCheck,
   FilterOperatorMapping,
+  FilterDisplayTypeMapping,
 } from '../interfaces';
 import ProcessModelSearch from './ProcessModelSearch';
 import ProcessInstanceReportSearch from './ProcessInstanceReportSearch';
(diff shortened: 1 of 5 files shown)


madhurrya:
@jasquat Can you please explain this a bit. Is this related to the time tracker?
Not clear about this 'This adds in the ability to turn off "with-tasks-for-me" from a report'

burnettk:
if you're on the "for me" tab of process instances and save a perspective, it will be locked to process instances "for me."  then, if you decide you want to see other peoples' process instances, you can go to the "All" tab of process instances, select the report (it will be just "for me" instances even though you are on "All" tab), then go to Advanced, and you can uncheck "Include tasks for me" (and save the perspective if you want).

burnettk:
in terms of the date/time display stuff, see https://timetracking.spiffworkflow.org/process-models/extensions:my-time-tracking-report/form/results-markdown.md for how we configure the table at https://timetracking.spiffworkflow.org/extensions/my-time-tracking-report

dinithihj:
Tested. Metadata can be formatted as DateTime/Duration.

image

dinithihj:
@burnettk I have tested "Include tasks for me" feature and it functions as described in your earlier comment. However, I am not clear on the exact purpose of the "Include tasks for me" option.

As mentioned in your comment, if you are in "For me" tab and save a  perspective, and then if you decide to see other peoples' process instances and click on All tab. Why it will display just "for me" instances even though you are on "All" tab. 

I am clicking on All tab as my requirement is to view other peoples' process instances and I will have to perform additional steps (such as clicking on Advance and unchecking "Include tasks for me") to view the desired instances under the "All" tab.

And I can always go to "For me" tab to see my instances - which produce the same result as when "Include tasks for me" is checked.

Please let me know if there is a specific requirement for  "Include tasks for me" which I didn't notice.

Issue #538, opened 2023-10-11
Add support for a KKV data store


jbirddog:
Due to the nature of reading/writing from data stores it would be advantageous to have a `KKV` data store which deferred the actual read/write until a top level key is specified. This could be achieved with low effort if functions were returned from the data store read/write operations. Another option is to add extensions to the arrows from data stores to/from tasks which would specify the top level key.

dinithihj:
Tested Environment: dev.app
Frontend Version: 289
Backend Version: 292
Comment: New kkv items are visible under the "Select Data Store" dropdown list.

![image](https://github.com/sartography/spiff-arena/assets/123297840/7d3d4566-9deb-4b66-a5f6-763788318992)

madhurrya:
Released to Prod on November 27th

usama9500:
Closing this issues as  documented KKV data store at https://github.com/sartography/spiff-arena/pull/1360

PR #548, opened 2023-10-16
Keyed key/value data store implementation

jbirddog:
Edit: this is for #538 

Data store implementation which supports two levels of keys. When the data store is written to, dictionaries are used such as:

```
datastore = {
  "k1": {
    "k2": "some updated value",  // add two different key/value pairs under "k1" 
    "k22": "value 2"
  }, 
  "k3": None,  // delete any values under the top level key "k3"
  "k4": {"k5": None},  // delete the value for "k5" under "k4"
  "k6": {"k7": "bob"}}  // add a single value for "k7" under "k6"
}
```

When reading from the data store, a getter function is placed into task data. which can be used to pull individual values:

```
x = datastore("k1", "k2")
y = datastore("k1", "k22")
z = datastore("k6", "k7")
no_value = datastore("k4", "k5")
```


spiffworkflow-backend/src/spiffworkflow_backend/data_stores/kkv.py
```diff
@@ -0,0 +1,87 @@
...
+from SpiffWorkflow.bpmn.specs.data_spec import BpmnDataStoreSpecification  # type: ignore
+from SpiffWorkflow.task import Task as SpiffTask  # type: ignore
+from spiffworkflow_backend.models.db import db
+from spiffworkflow_backend.models.kkv_data_store import KKVDataStoreModel
+
+
+class KKVDataStore(BpmnDataStoreSpecification):  # type: ignore
+    """KKVDataStore."""
+
+    def _get_model(self, top_level_key: str, secondary_key: str) -> KKVDataStoreModel | None:
```
burnettk, line 13: am i reading correctly that the database columns for k1 and k2 are allowing null but these keys are needing to be strings? is there an issue here?

spiffworkflow-backend/src/spiffworkflow_backend/models/kkv_data_store.py
```diff
@@ -0,0 +1,16 @@
...
+
+@dataclass
+class KKVDataStoreModel(SpiffworkflowBaseDBModel):
+    __tablename__ = "kkv_data_store"
+
+    id: int = db.Column(db.Integer, primary_key=True)
+    top_level_key: str = db.Column(db.String(255), index=True)
+    secondary_key: str = db.Column(db.String(255), index=True)
+    value: dict = db.Column(db.JSON)
+    updated_at_in_seconds: int = db.Column(db.Integer)
```
burnettk, line 15: i think jason and i decided that these blah_in_seconds columns should always be set and the columns should therefore be `nullable=False`. it's configured wrong in almost every model, but we might as well start correctly here.

burnettk (approved):

spiffworkflow-backend/src/spiffworkflow_backend/data_stores/kkv.py
jbirddog: Good catch, updated.

Issue #594, opened 2023-10-26
First 5 Minutes


danfunk:
The first 5 minutes someone uses SpiffWorkflow Application can be very daunting.  Here are issues I've worked through with several new users, or suggested by team members, that should be addressed:

**General Features:**
1. We should display directions when viewing an empty Progress Group that you can edit  ("This is an empty process Group.  To get started you should create a new Process Model using the button below") ✔️ 
1. If you can't start a process model because it doesn't have a BPMN file, don't show a start button. ✔️ 
1. If you are viewing a process model and it doesn't have any files in it - Ask the user if they would like to create a new BPMN file. ✔️
1.  #797 ✔️ 

Use cases to capture imagination of users:
 * integrate with openai (respond to prompt, generate image)
 * read and write to database
 * send discord message
 * fire http request to example data endpoint
 * generate "if statement"

Figma ideas [here](https://www.figma.com/file/uLnj6IoyGxBY68hd8Kpu8X/First-five-minutes-palette-ideas?type=whiteboard&node-id=0%3A1&t=lDHKGELgWqeKVaz5-1). Ayoub is sketching out a [proposal](https://ayoub98.notion.site/BPMN-New-Fresh-UX-UI-fa104d6e0345431ca565bcf4db5cf693).

jasquat:
Draft PR at https://github.com/sartography/spiff-arena/pull/601

dinithihj:
Tested Environment: dev.app
Frontend Version: 296
Backend Version: 299

1. Directions displayed for empty process models, However directions are still available even if the user does not have permission to add process models/groups, which is not correct.
![image](https://github.com/sartography/spiff-arena/assets/123297840/46a1fb27-13ca-4e31-9d4e-d170bd446c52)
![image](https://github.com/sartography/spiff-arena/assets/123297840/f6bf2371-fb6b-4bd0-810a-58aff5ce625c)

2. & 3. The start button is not visible when there's no BPMN file associated with it. And directions given on how to create BPMN file

![image](https://github.com/sartography/spiff-arena/assets/123297840/dca3507a-2f3f-484e-803c-07b4177044b9)
![image](https://github.com/sartography/spiff-arena/assets/123297840/a7328ff6-c73d-43b7-9872-6057715c3be4)

jasquat:
Removed instructions to add a process model / group if a user does not have access. It now just says there aren't any to display.

dinithihj:
Tested Environment: dev.app
[...]

PR #601, opened 2023-11-01
Feature/first five minutes

jasquat:
This implements #594. It adds some instructions of what to do if no process models or groups are found or a model doesn't have any files. It also hides the Start button if a model doesn't have a primary bpmn file.


danfunk:
so much love for this.

Issue #563, opened 2023-10-23
Design CRUD support for data stores


jbirddog:
Need to decide how we are going to support CRUD operations across multiple data store types, ideally in some common interface. One approach could be an interface beyond that exposed by SpiffWorkflow that allows each data store to describe how to add/update/delete etc records with its realm. Once this is designed out separate tickets will likely be required for the implementation for the various data stores.

dinithihj:
@jbirddog Can you please let me know whether is there anything to test related to this ticket?

jbirddog:
This was more of an internal code change in the backend, no new functionality was added. Everything should work as before with regards to the data store tab and interaction with data stores in the diagrams.

dinithihj:
Executed all tests and everything looks ok.

dinithihj:
Released to Prod on December 21st

PR #689, opened 2023-11-20
Data Store CRUD interface

jbirddog:
Work on #563 - this adds a generic interface that classes can implement which would allow any compliant data store to be interacted with from the data store viewer in a generic fashion. This is only partially implemented ~~(just the kkv data store, discovery of all data stores is not defined for the drop down, etc). Since the ticket is really just about defining the interface I didn't want to take it too far, but did want one implementation as an example.~~ only the read operations are supported for the model based data stores (JSONFileDataStore is omitted). Since these data stores adhere to the new interface we can now start doing basic CRUD operations across them all in a uniform fashion.


spiffworkflow-backend/src/spiffworkflow_backend/data_stores/crud.py
```diff
@@ -0,0 +1,27 @@
+from typing import Any
+
+
+class DataStoreCRUD:
+    @staticmethod
+    def existing_data_stores() -> list[dict[str, Any]]:
+        raise Exception("must implement")
+
+    @staticmethod
```
burnettk, line 9: if used the word data store in a very generic way, such that arbitrary database tables with arbitrary properties could be included, the only thing i can see in this interface that wouldn't fit very well is that query_data_store doesn't accept all of the querying options you might want, such as filtering by column values and sorting. that being said, perhaps you wouldn't be allowed to get that fancy via this "API." it seems that you have the same limitation with kkv, right, where you can get stuff out, but you have to get everything? you can't filter by secondary key via this "API"? Perhaps there is another way you can filter by secondary key, and a similar thing could be used for these "data model data stores" that i'm hoping we build?

burnettk (approved):

spiffworkflow-backend/src/spiffworkflow_backend/data_stores/crud.py
jbirddog: we just discussed and all agree the idea is good, going to defer and do the basic crud operations first.

PR #769, opened 2023-11-28
reset to page 1 when status changes to fix #765

### LINKED ISSUE #765: Bug in process instances filter
When moving to the next pages and applying a new filter with a reduced number of results, no results will be visible in the results section.

For instance, consider a scenario where the user is in the "My Process Instances" tab. Upon clicking the "Next page" button, the user moves to page 4. While on page 4, the user applies another filter that returns only 14 records. Consequently, no data is presented in the "My Process Instance" tab, leading to confusion. This issue arises because the user was previously on page 4, and presently, there is insufficient data to populate page 4.

Image

Note that if the user returns to page 1 by clicking the "Previous Page" option or selecting 1 from the page number list, the results of the new filter become visible.

Link to the Video - https://drive.google.com/file/d/1J8YUdIfwbnx4Hm4PPKCFwHmhMK3cKvo1/view?usp=sharing

burnettk:
Reset to page 1 if the user changes a status filter.

### DIFF
--- a/spiffworkflow-backend/.snyk
+++ b/spiffworkflow-backend/.snyk
@@ -2,26 +2,12 @@
 version: v1.25.0
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
 ignore:
+  # in case snyk wants werkzeug v3, in the future:
   # we cannot upgrade werkzeug because it breaks connexion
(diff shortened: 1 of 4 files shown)


burnettk:
fixes #765

burnettk:
closes https://github.com/sartography/spiff-arena/issues/765

PR #799, opened 2023-12-11
Numeric range field added

Kayvon-Martinez:
Added a custom field to handle numeric ranges. The field checks for the first key in the properties of the schema and will add min and max to it when the values are put in. The field will check if the min value is more than the max and swap the values accordingly, allows the input of decimal values, and inserts commas where they should be in the input. The values are stored in the first object in the properties.


burnettk:
merged it and filed new PR against main: https://github.com/sartography/spiff-arena/pull/800. next time feel free to file the PR against main directly.

Kayvon-Martinez:
@burnettk Okay I will. Thank you.

PR #925, opened 2024-01-29
Added vertical option to radio buttons and fixed the four problems with the compensation range field

Kayvon-Martinez:
Added support for a "column" property that takes a true or false value for radio buttons. If the value is true the radio buttons will be vertical and horizontal if false or unspecified.
image
Fixed the issue of not being able to input decimal values in the numeric range field, not being able to input a hyphen the first time, not being able to input a zero the first time, and not getting a validation error after removing the min value when the field is required. There are no new changes to how the schema should work for the numeric range field.
image

## Summary by CodeRabbit

- **New Features**
	- Added a script to retrieve BPMN JSON representation for a given process instance.
	- Introduced a new BPMN process for fetching and displaying random facts.
- **Bug Fixes**
	- Fixed workflow schedule for deleting old container images and adjusted retention policy.
	- Removed unused imports and functions across various services for cleaner operations.
- **Documentation**
	- Updated and clarified the explanation for displaying fields side by side in forms.
	- Removed sections detailing configuration of Escalation Events Properties and fast feedback on errors with Python expressions.
- **Refactor**
	- Simplified data store operations and message sending logic by removing unnecessary parameters and logic.
	- Streamlined authentication and authorization services for more efficient token handling and access control.
- **Chores**
	- Downgraded several dependencies to improve compatibility and stability.
	- Removed the `openai` dependency and updated several package versions in `pyproject.toml` files.
- **Revert**
	- Reverted the version of `poetry` in various configurations for better project compatibility.


burnettk:
did you consider a "ui:" prefix on the "column" key? i'm not sure what is better here, with or without.

spiffworkflow-frontend/src/rjsf/custom_widgets/NumericRangeField/NumericRangeField.tsx
```diff
@@ -80,11 +96,38 @@ export default function NumericRangeField({
...
+      setMinValue(formatNumberString(numberValue.toString()));
+    }
+    if (nameToChange === 'max') {
+      max = numberValue;
+      setMaxValue(formatNumberString(numberValue.toString()));
+    }
     if (!disabled && !readonly) {
       if (nameToChange === 'min' && numberValue > max) {
+        min = numberValue;
+        setMinValue(numberValue.toString());
```
burnettk, line 128: i think we already called `setMinValue` above in this case on line 119. one place it has `formatNumberString` and one place it doesn't, but i'm not sure which is correct.

burnettk (approved):

burnettk:
@harmeet-status sure, that naming would work for me. maybe we let kayvon make the final call now that we've made the decision more difficult by adding more options. :)

harmeet-status:
> @burnettk I was thinking how about you do this after talking to @Kayvon-Martinez:
> 
> `"ui:layout": column/row` This would make it more easily readable. Default value is row if not defined.

@Kayvon-Martinez lets go with this, @burnettk has given his blessings

Kayvon-Martinez:
I have changed the "column" to "ui:layout" and it now accepts "column" and "row" as values, and will default to row. I have also fixed the decimal support for the compensation range field. The formatting should work fine with decimals now and there are no changes needed to be made to the schema.

Kayvon-Martinez:
I have added a character counter field that takes a "maxLength" value for the limit on the amount of characters that can be entered. Here are the example schemas.
jsonSchema
image
uiSchema
image

I have also changed the numeric range field so that it has specific error messages depending on what values are passed in for the min and the max. There are no changes that need to be made to the schema for the numeric range field.

PR #949, opened 2024-02-02
Add data stores API

theaubmov:


burnettk (approved):
tests failed with:

[eslint] 
src/components/ReactDiagramEditor.tsx
  Line 420:6:  React Hook useEffect has a missing dependency: 'onDataStoresRequested'. Either include it or remove the dependency array. If 'onDataStoresRequested' changes too often, find the parent component that defines it and wrap that definition in useCallback  react-hooks/exhaustive-deps

otherwise seems good.

PR #1159, opened 2024-03-04
Error Events Documentation

usama9500:
Created Documentation on:

- Error Events
- Types of Error Events
- Example of Each Error Event

## Summary by CodeRabbit

- **Documentation**
	- Enhanced documentation on Error Events in BPMN, focusing on the Error Start Event for initiating error handling subprocesses in business process workflows.
	- Updated the document list in `docs/index.md` to include information about Error Events.


burnettk:
@usama9500 , please check the suggestions from coderabbit.

usama9500:
> @usama9500 , please check the suggestions from coderabbit.

Updated.

danfunk:
This is a good general overview of what Errors are, but it doesn't capture how to use them within SpiffWorkflow - I think we need to add some examples to the SpiffDemo site that acutally use errors.  I'm sure that Alex or Madhurya could help define these, and we could link to them from the documentation.  Then people could really see them working.  Connecting this documentation to real usage within SpiffWorkflow is critical to establishing a good reputation.  Otherwise people will start to suspect that these docs are generic and not really about our product.

usama9500:
> This is a good general overview of what Errors are, but it doesn't capture how to use them within SpiffWorkflow - I think we need to add some examples to the SpiffDemo site that acutally use errors. I'm sure that Alex or Madhurya could help define these, and we could link to them from the documentation. Then people could really see them working. Connecting this documentation to real usage within SpiffWorkflow is critical to establishing a good reputation. Otherwise people will start to suspect that these docs are generic and not really about our product.

Thanks Dan for the feedback. 
Yes, that's the plan. Basically there wasn't any documentation on error events previously. So I decided to first cover basics and then add examples. Since there is only one example on error events, I am talking with Alex and Madhurya to provide me some running examples so I can add it.

danfunk (approved):

PR #1204, opened 2024-03-14
Adding Example of Signal Events and New documentation on Multiinstanc…

usama9500:
Added example of:
- Signal Events

Created documentation and example for:
- Multiinstance tasks 

## Summary by CodeRabbit

- **Documentation**
	- Introduced multi-instance tasks in BPMN with sequential and parallel execution modes, covering key properties and providing practical examples.
	- Detailed the usage of Signal Boundary Events in BPMN processes with examples of signal-based routing and timed controls.
	- Updated documentation index to reference the new guides on multi-instance tasks and signal events.
	- Added a new entry `.aider*` to the `.gitignore` file.


docs/Building_Diagrams/signal_events.md
```diff
@@ -74,3 +74,55 @@ In an online shopping system, when a customer's payment is successfully processe
...
+
+#### **Signal Boundary Event 1**: 
+Catches the signal for "eat_spam" and redirects the workflow to a Manual Task named "Spam Message".
+
+![signal_event_example](images/Signal_events_spiff_example3.png) 
+
+#### **Signal Boundary Event 2**: 
+Listens for the "eat_cheetos" signal, leading the process to a Manual Task labeled "Cheetos Message".
+
+![signal_event_example](images/Signal_events_spiff_example4.png)
```
burnettk, line 106: you end up with about four images that basically show the same diagram, but with different elements selected. i think just one picture of the diagram would be sufficient. it might be nice to show a picture of the form so people can get an idea what the signal buttons look like when you run the process?

docs/Building_Diagrams/multiinstance.md
```diff
@@ -0,0 +1,110 @@
...
+       {"composer": "Ludwig van Beethoven", "genre": "Classical/Romantic"},
+       {"composer": "Wolfgang Amadeus Mozart", "genre": "Classical"}
+     ]
+```
+
+#### 3. **Multi-Instance Task - Edit Composer**:
+
+This task is configured as a parallel multi-instance task that iterates over the `composers` array created by the previous script task. It allows for the editing of each composers information within the array.
+
+![Multi_instance_example](images/multiinstance_example3.png)
```
burnettk, line 75: same here; i'd probably only use the image once. the highlighting is helpful, but i don't think it's helpful enough. not sure what others might think.

burnettk: also, if the picture were bigger (cropped to just the tasks), it would be easier to read the activity names and match them up to the descriptions you're giving.

burnettk (approved):

docs/Building_Diagrams/signal_events.md
usama9500: Sure thing, I will update it. Basically I added step by step images to display how properties panel show for each element.

PR #1231, opened 2024-03-18
Fix warning during pre-commit

jbirddog:
Warning was `Unexpected key(s) present on local => black: line-length` believe this is safe since `--line-length` is passed in args already but have _edit: not_ rigorously tested. The entire codebase did not reformat however so i take that as a good sign. Also removed a couple commented out lines while i was in there.

## Summary by CodeRabbit

- **Chores**
	- Updated pre-commit hook configurations for improved code formatting and compliance.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,14 +3,11 @@ repos:
     hooks:
       - id: black
         name: black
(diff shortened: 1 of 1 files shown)


jasquat (approved):
Seems good. Looks like the args a few lines down were supposed to handle this anyway.

burnettk (approved):

PR #1263, opened 2024-03-23
Implement React Query (Tanner)

tcoz:
Also, replace PermissionCache with the framework's hooks.

## Summary by CodeRabbit

- **New Features**
	- Integrated `@tanstack/react-query` for enhanced data fetching and state management.
	- Added `ReactQueryDevtools` for better debugging of query states.
- **Refactor**
	- Refactored permission service logic for more efficient permission checking and rule updating.
- **Chores**
	- Updated project dependencies to include new tools for linting and development efficiency.


burnettk:
FYI, cypress tests are broken on main, so be aware that your branch won’t pass. There are two commits that might have broken it, and it doesn’t seem to fail locally, so it’s a little less straightforward to fix than normal, which is why we didn’t hammer it yesterday. We’ll get it back in action asap.

tcoz:
Ah good, I saw no earthly reason why my changes would effect a scroll window in Cypress.

burnettk (approved):

PR #1340, opened 2024-04-05
allow accessing backend at either /api or /

burnettk:
this hooks into connexion and the openid blueprint, mounting both at / AND /api


mogul:
This is interesting! Is there a reason to try to infer the path the request came from, rather than take it from a configuration variable? (I thought this was what the existing `APPLICATION_ROOT` or `SPIFFWORKFLOW_BACKEND_APPLICATION_ROOT` variables were for, but I haven't seen that changing them does what I expect.)

burnettk:
thank you for the question about APPLICATION_ROOT. that lead us towards a better solution: https://github.com/sartography/spiff-arena/pull/1344.

we ended up using the SCRIPT_NAME feature built in to the wsgi standard, as described by this annoyed person: https://dlukes.github.io/flask-wsgi-url-prefix.html

ultimately, i think if `/api` is making it past the reverse proxy and into the backend container, `SPIFFWORKFLOW_BACKEND_WSGI_PATH_PREFIX=/api` will work great once #1344 is merged.

burnettk:
we went with the https://github.com/sartography/spiff-arena/pull/1344 approach, and that change is now merged and ready to test out.

mogul:
Excellent, we'll rebase and give it a try! Thanks so much for making this easier.

mogul:
Not working... 😞 I'll put details on the other PR.

PR #1344, opened 2024-04-05
Allow setting path prefix

jasquat:
This PR removes support for SPIFFWORKFLOW_BACKEND_APPLICATION_ROOT in favor of SPIFFWORKFLOW_BACKEND_WSGI_PATH_PREFIX. This is useful if using path-based routing with the backend. This also removes the need to strip out the prefix before sending it to the backend.

## Summary by CodeRabbit

- **Refactor**
	- Updated environment variable names and usage for better clarity and consistency across the application.
- **Documentation**
	- Enhanced guidance on setting up path-based routing for running frontend and backend on the same host.
- **Chores**
	- Simplified the development and deployment process by streamlining environment variable settings.

PR #1349, opened 2024-04-07
Prevent duplicated path prefix elements in generated URLs

mogul:
Refer to [this comment](https://github.com/sartography/spiff-arena/pull/1344#issuecomment-2041529243) for the context. Repeating the gist of it:

> Throughout this file, `host_url` is set in a line like:
> 
> ```python
>  host_url = current_app.config.get("SPIFFWORKFLOW_BACKEND_URL") or request.host_url.strip("/")
> ```
> 
> ...and then in code below, it's used to generate URLs like
> 
> ```python
> f"{host_url}{url_for(ROUTENAME)}"
> ```
> 
> 1. `SPIFFWORKFLOW_BACKEND_URL` _must_ be set for various other parts of the app to work, and it's operator-configured. There are two valid cases we're concerned with:
>    a. The URL ends with a path element
>    b. The URL does not end with a path element
> 
> 2. `request.host_url` will _never_ end in a path element because it stops at the hostname by design.
> 
> 3. `url_for()` will _always_ prepend the WSGI-provided path element (eg `/flarble`) for the containing app.
> 
> Consider the case where `SPIFFWORKFLOW_BACKEND_URL` ends in `/flarble`, and `SPIFFWORKFLOW_BACKEND_WSGI_PATH_PREFIX` is also `/flarble`. The combination of `1.a` and `3` will generate a bad URL (eg `/flarble/flarble/ROUTENAME`).

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/routes/openid_blueprint/openid_blueprint.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/routes/openid_blueprint/openid_blueprint.py
@@ -42,7 +42,7 @@ def well_known() -> dict:
     """
 
     # using or instead of setting a default so we can set the env var to None in tests and this will still work
-    host_url = current_app.config.get("SPIFFWORKFLOW_BACKEND_URL") or request.host_url.strip("/")
+    host_url = current_app.config.get("SPIFFWORKFLOW_BACKEND_URL").removesuffix(request.root_path) or request.host_url.strip("/")
     return {
         "issuer": f"{host_url}/openid",
         "authorization_endpoint": f"{host_url}{url_for('openid.auth')}",
@@ -55,7 +55,7 @@ def well_known() -> dict:
 @openid_blueprint.route("/auth", methods=["GET"])
 def auth() -> str:
     """Accepts a series of parameters."""
-    host_url = current_app.config.get("SPIFFWORKFLOW_BACKEND_URL") or request.host_url.strip("/")
+    host_url = current_app.config.get("SPIFFWORKFLOW_BACKEND_URL").removesuffix(request.root_path) or request.host_url.strip("/")
     return render_template(
         "login.html",
         state=request.args.get("state"),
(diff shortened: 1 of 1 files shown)


spiffworkflow-backend/src/spiffworkflow_backend/routes/openid_blueprint/openid_blueprint.py
```diff
@@ -114,13 +114,12 @@ def token() -> Response | dict:
     authorization = base64.b64decode(authorization).decode("utf-8")
     client_id = authorization.split(":")
 
-    host_url = current_app.config.get("SPIFFWORKFLOW_BACKEND_URL", request.host_url.strip("/"))
-    base_url = f"{host_url}/openid"
```
mogul, line 118: I removed the intermediate `base_url` here because I wanted the token's issuer value to exactly match the issuer specified in line 47 even under casual observation.

mogul:
That's failing tests because it loses the port where the app is running, if it was provided.🤔

OK, new plan: We'll use `SPIFFWORKFLOW_BACKEND_URL`, but we'll trim off the [`root_path`](https://flask.palletsprojects.com/en/2.3.x/api/#flask.Request.root_path).

burnettk:
local url debug info with and without `SPIFFWORKFLOW_BACKEND_WSGI_PATH_PREFIX=/api`:

```sh
> curl "http://localhost:7000/api/v1.0/debug/url-info"
{"request.root_path":"/api","request.host_url":"http://localhost:7000/","request.url":"http://localhost:7000/api/v1.0/debug/url-info","cache":{}}

> curl "http://localhost:7000/v1.0/debug/url-info"
{"request.root_path":"","request.host_url":"http://localhost:7000/","request.url":"http://localhost:7000/v1.0/debug/url-info","cache":{}}
```

burnettk:
this is awesome. moved party to a new branch that builds on these commits over at https://github.com/sartography/spiff-arena/pull/1350.

regarding "That's failing tests because it loses the port where the app is running, if it was provided", it would never lose the port in real life, even in development mode, and since request.host_url is sufficient, I think we should go with your first suggestion and avoid config where possible. i made that change over there.

Issue #773, opened 2023-11-30
Old instances BO/SME approve - time out error/slowness


harmeet-status:
UPDATE 20240312: proposal: avoid handing completed subprocesses to spiff to improve performance.

Noticed that instances create about 10 day or ago (in [test.app](http://test.app/) and [test.mo](http://test.mo/)d both) when BO/SME approves it waits in that page for a long time and shows an error message. Please see the screenshot.
For the instances not that old, didn’t notice the error, but some of them also take a long time to complete in the above step(may be the older ones get more time??)
Here are some videos https://drive.google.com/drive/folders/1bBliE4uYO3a2CjGam3ecdmyUls0bsp8v?usp=sharing

Even if we get these errors, it seems it gets assigned to the next reviewer correctly and they can continue the process.

![Image](https://github.com/sartography/spiff-arena/assets/104047049/4b5a389e-994f-4533-afcd-83265a5f28e5)

![Image](https://github.com/sartography/spiff-arena/assets/104047049/3e77814f-5f68-4b8d-80e3-4829f7e1c70e)

harmeet-status:
Comment from @madhurrya:

Update : when testing old instances for the current release in test.mod I am still noticing this behaviour. I didn’t get any errors this time, but it seems when the instance is older it take more time when submitting the form after BO/SME reviewing. (noticing this with about 20-30 days old instances)

madhurrya:
I am still noticing the delay as the instance get older.

For example please check this video. It shows 2 instances, old one created about 12 days ago and the new one created 1 day ago.
https://drive.google.com/file/d/1Eaw9yJhpLIM4jGMcRMHnklOvCYlCG5Nb/view?usp=sharing

The older it gets, the time increases.
@harmeet-status do we need to fix this?

harmeet-status:
Yes we need to fix performance issues. Back before we went Live, we spent a fair amount of effort focusing on performance and I think that we should be thinking now that we need to scale, so everything that we build has to also be performant!

harmeet-status:
What work is remaining on this?

burnettk:
we think we should add an optimization to avoid considering completed subprocesses. we'll try that and see if it is implementable.

jasquat:
Note we will wait for https://github.com/sartography/SpiffWorkflow/issues/395 before attempting anything with this ticket.

jasquat:
This is ready for qa on dev.mod.

madhurrya:
[...]

PR #1370, opened 2024-04-09
Look for bad get_current_user scripts

burnettk:
also puts xml parsing in one place and makes it a little safer.
also uses ruff check to avoid deprecation of `ruff .`
deals with comment on #773 about using get_current_user breaking some instances.

## Summary by CodeRabbit

- **New Features**
	- Introduced a script to identify uses of `get_current_user()` in BPMN files, enhancing security and compliance checks.

- **Refactor**
	- Improved the autofix process by integrating `check` with `--fix` in the `Makefile` and `run_pyl` script for more comprehensive code quality checks.
	- Enhanced XML parsing security and efficiency in the process model testing service by updating method implementations for better file handling and parser configuration.

- **Chores**
	- Updated internal scripts and services for better performance and security compliance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ pre-commit:
 	$(IN_ARENA) poetry run pre-commit run --verbose --all-files
 
 ruff:
-	$(IN_ARENA) poetry run ruff --fix spiffworkflow-backend
+	$(IN_ARENA) poetry run ruff check --fix spiffworkflow-backend
 
 run-pyl: fe-lint-fix ruff pre-commit be-mypy be-tests-par
 	@true
--- a/bin/run_pyl
+++ b/bin/run_pyl
@@ -24,7 +24,7 @@ function get_python_dirs() {
 function run_autofixers() {
   python_dirs="$(get_python_dirs) bin"
   # shellcheck disable=2086
-  poetry run ruff --fix $python_dirs || echo ''
+  poetry run ruff check --fix $python_dirs || echo ''
 }
 
 function run_pre_commmit() {
--- a/spiffworkflow-backend/bin/lint_get_current_user_scripts.py
+++ b/spiffworkflow-backend/bin/lint_get_current_user_scripts.py
@@ -25,7 +25,6 @@ def check_script_and_prescript_elements(tree, bpmn_file_path: str) -> None:
     # Find all script tasks and preScript elements
     script_tasks = tree.xpath("//bpmn:scriptTask", namespaces=nsmap)
     pre_scripts = tree.xpath("//spiffworkflow:preScript", namespaces=nsmap)
-    post_scripts = tree.xpath("//spiffworkflow:postScript", namespaces=nsmap)
 
     # Check script tasks for get_current_user() calls
     for task in script_tasks:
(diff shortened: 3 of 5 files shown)


jbirddog (approved):

madhurrya:
@burnettk is this deployed in dev.mod?

burnettk:
there is nothing to deploy, since the script is run locally. if we want to know if there are bad scripts on a certain branch, i can check.

madhurrya:
I am not clear KB. Doesn't these files related to this PR needs to be deployed in dev.mod if I need to check if this fixes the old instance issue I have mentioned in #773?

burnettk:
@madhurrya there is no reasonable way to fix instances that were run with a process model that doesn't conform to these policies. ideally we fix process models so no new bad instances are created, flush out the bad instances until we feel we can manage the rest manually, and then deploy celery.

burnettk:
we could theoretically write a script to figure out if any open process instances would be affected. it would be significantly harder to write than the existing script, and even more difficult on prod, where we don't have access.

madhurrya:
If we update the existing models accordingly now, then we'll not get this issue with the old instances of those updated models right? Have we already updated those models or not yet? Have you already discussed this with @sashayar13 ?

sashayar13:
I haven't updated all process models yet, was planning to do so during the week. But it will not solve the issue - there still will be old open instances in prod since people are doing the expense submission to the previously opened Travel and Purchase requests. We need to have a way to handle such old instances

burnettk:
ok, first we need to fix the models in prod, and then i see a few options here:

1. note the time that we updated prod, maybe 30 apr 2024, or whenever that happens. wait a few weeks. do a query for instances created before 30 apr 2024. those instances are bad. if there are 3, talk to the the people who "own" the instances, mitigate the potential impact, and deploy celery. if there are 50, wait a few more weeks. eventually there are few enough instances that we can safely release celery without causing too much damage.
2. write a script that detects bad instances. this would take about 3 or 4 days. that way we could have a better idea about which of the 50 instances could actually be affected by the get_current_user issue. this is not the greatest type of work because part of it would only be applicable for this one issue, and then we'd throw it away.
3. try to really understand the nature of the process models that are affected and figure out a way to make old instances work by writing some sort of script. this is likely to be a 5 to 10 day task, most of the value of which wouldn't be carried forward (the learnings and code produced would only be applicable for this one issue, and then we'd throw it away), so i wouldn't recommend it.

sashayar13:
@burnettk, we have already updated all process models affected by the change. These process models are in prod from 29.04

I can check how many open instances we have created before this date and let you know. Based on the number and nature we can see which option is acceptable

sashayar13:
373 not completed process instances with the start date before 29.04.2024
Next step - analyse the nature of the process instances and plan the next steps accordingly

madhurrya:
Since we are still testing and bug fixing things in Celery I assume it'll take atleast another month or so to reach it to Prod. By that time wouldn't most of those instances will be closed?
Still we'll have to take action for any instances that'll be open.

sashayar13:
@madhurrya, may I ask you please to inform me a week prior to the change release date?

madhurrya:
@sashayar13 I guess you mean before the Celery release. Yes sure will let you know.

PR #1375, opened 2024-04-10
Attempt to detect unused code in the frontend

jbirddog:
Trying out some frontend libs to detect unused code, seems to be a lot of false positives but this one didn't show up in a grep.


burnettk (approved):
yeah, i think this is legit, thanks. we moved from tasks to instances at some point, and we were tentative about removing working code, but it's time.

burnettk:
this is super hot, nice find. we made `bin/codemod/remove_all_unused_functions` in backend, but i'm pretty sure it doesn't work. :)

jbirddog:
> this is super hot, nice find. we made `bin/codemod/remove_all_unused_functions` in backend, but i'm pretty sure it doesn't work. :)

very cool, had no idea that existed, i'll give it a try and see if it returns anything useful.

PR #1403, opened 2024-04-15
Pi file data on fs

jasquat:
Supports saving the process instance data file on the file system in a hashed directory structure.

Setting the config `SPIFFWORKFLOW_BACKEND_PROCESS_INSTANCE_FILE_DATA_FILESYSTEM_PATH` will enable saving these files to where this config is pointing.

## Summary by CodeRabbit

- **New Features**
    - Refactored database configuration for improved readability.
    - Added functionality to store files associated with process instances.
    - Implemented file handling functions for storing and retrieving files on the file system.
    - Conditional check to retrieve file contents from the file system.
    - Introduced a class for migrating process instance file data from the database to the file system.

- **Bug Fixes**
    - Updated file data handling in process instance services.

- **Tests**
    - Added test scenarios for downloading files from the file system.
    - Validation of directory structure generation for process instance file data.
    - Test case for migrating process instance file data to the file system.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/models/process_instance_file_data.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/models/process_instance_file_data.py
@@ -1,11 +1,15 @@
+import os
 from dataclasses import dataclass
 
+from flask import current_app
 from sqlalchemy import ForeignKey
 from sqlalchemy.dialects.mysql import LONGBLOB
 
 from spiffworkflow_backend.models.db import SpiffworkflowBaseDBModel
 from spiffworkflow_backend.models.db import db
 from spiffworkflow_backend.models.process_instance import ProcessInstanceModel
+
+PROCESS_INSTANCE_DATA_FILE_ON_FILE_SYSTEM = "contents:in:filesystem"
 
 
 @dataclass
@@ -17,7 +21,20 @@ class ProcessInstanceFileDataModel(SpiffworkflowBaseDBModel):
     mimetype: str = db.Column(db.String(255), nullable=False)
     filename: str = db.Column(db.String(255), nullable=False)
     # this is not deferred because there is no reason to query this model if you do not want the contents
-    contents: str = db.Column(db.LargeBinary().with_variant(LONGBLOB, "mysql"), nullable=False)
+    contents: bytes = db.Column(db.LargeBinary().with_variant(LONGBLOB, "mysql"), nullable=False)
     digest: str = db.Column(db.String(64), nullable=False, index=True)
     updated_at_in_seconds: int = db.Column(db.Integer, nullable=False)
     created_at_in_seconds: int = db.Column(db.Integer, nullable=False)
+
+    def store_file_on_file_system(self) -> None:
+        filename = self.digest
+        filepath = os.path.join(current_app.config["SPIFFWORKFLOW_BACKEND_PROCESS_INSTANCE_FILE_DATA_FILESYSTEM_PATH"], filename)
+        with open(filepath, "w") as f:
+            f.write(self.contents.decode("utf-8"))
+        self.contents = PROCESS_INSTANCE_DATA_FILE_ON_FILE_SYSTEM.encode()
+
+    def get_contents_on_file_system(self) -> bytes:
+        filename = self.digest
+        filepath = os.path.join(current_app.config["SPIFFWORKFLOW_BACKEND_PROCESS_INSTANCE_FILE_DATA_FILESYSTEM_PATH"], filename)
+        with open(filepath) as f:
+            return f.read().encode()
--- a/spiffworkflow-backend/src/spiffworkflow_backend/models/process_instance_file_data.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/models/process_instance_file_data.py
@@ -10,6 +10,7 @@ from spiffworkflow_backend.models.db import db
 from spiffworkflow_backend.models.process_instance import ProcessInstanceModel
 
 PROCESS_INSTANCE_DATA_FILE_ON_FILE_SYSTEM = "contents:in:filesystem"
+PROCESS_INSTANCE_DATA_FILE_ON_FILE_SYSTEM_DIR_COUNT = 2
 
 
 @dataclass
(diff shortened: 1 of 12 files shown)


jbirddog (approved):

spiffworkflow-backend/src/spiffworkflow_backend/services/process_instance_service.py
```diff
@@ -443,6 +443,8 @@ def save_file_data_and_replace_with_digest_references(
         models = cls.replace_file_data_with_digest_references(data, process_instance_id)
 
         for model in models:
+            if current_app.config["SPIFFWORKFLOW_BACKEND_PROCESS_INSTANCE_FILE_DATA_FILESYSTEM_PATH"] is not None:
+                model.store_file_on_file_system()
```
jakubgs, line 447: I see, so you want to retain the old method and keep it as default?

I guess it's a good way to test things out, but ultimately I would advise against keeping file storage in the database as a sane default, it's not sane =].

spiffworkflow-backend/src/spiffworkflow_backend/models/process_instance_file_data.py
```diff
@@ -17,7 +22,40 @@ class ProcessInstanceFileDataModel(SpiffworkflowBaseDBModel):
...
+
+    def get_contents_on_file_system(self) -> bytes:
+        filepath = self.get_full_filepath()
+        with open(filepath) as f:
+            return f.read().encode()
+
+    def get_full_filepath(self) -> str:
+        dir_parts = self.__class__.get_hashed_directory_structure(self.digest)
+        return os.path.join(
+            current_app.config["SPIFFWORKFLOW_BACKEND_PROCESS_INSTANCE_FILE_DATA_FILESYSTEM_PATH"], *dir_parts, self.digest
```
jakubgs, line 51: Isn't `self.digest` the whole thing? I thought you were trying to reproduce a `Git` type structure where the folder structure is subtracted from the file, so the whole path is the full digest:
```
~/work/spiff-arena deploy-mod-dev
 > ls -l .git/objects/01/fd9e732b52acf6b310e54d1490c8ea1e43f811 
-r--r--r-- 1 jakubgs jakubgs 1388 Feb  7  2023 .git/objects/01/fd9e732b52acf6b310e54d1490c8ea1e43f811

~/work/spiff-arena deploy-mod-dev
 > git cat-file -t 01fd9e732b52acf6b310e54d1490c8ea1e43f811 
blob
```
But it seems like here that would be more like `01/01fd9e732b52acf6b310e54d1490c8ea1e43f811`, which is weird.

jakubgs (approved):
Thank you. I guess I'll need to develop a migration script to handle the switch too?

burnettk:
> Thank you. I guess I'll need to develop a migration script to handle the switch too?

if you’re willing. Maybe we could run it at app boot if we detect that the first row in the file data table has a big contents or similar.

spiffworkflow-backend/src/spiffworkflow_backend/models/process_instance_file_data.py
burnettk: Yeah, that’s interesting. I went looking for a standard for this and it eluded me. Git’s implementation is definitely an important data point. I like that it is not duplicated in that approach.

I like using the full Digest because you can copy and paste it and slap it into a find –name, so you can very easily restructure the number of prefixed directories, or even attach a file to a GitHub issue without renaming and it remains coherent.

@jbirddog @jasquat ?

jakubgs: The point about finding it is a good one, but it seems to me like manual management of those files would be an edge case, more so reserved for developers, less so for administrator and not at all for users. But it makes sense in that regard.

jakubgs:
Hm, this would be actually more tricky than I thought, since we would not just extract data from the DB but also modify the DB to point to the location correctly, so this becomes more so like an actual full blown migration rather than just a one-off admin task.

Do you think this should be a proper migration? This kinda depends on whether you want to retain DB file storage as the default - which I would consider a poor choice - or whether you'd make the filesystem solution the default. If the latter this can be more of a quick-and-dirty script to do it 5 times on our deployment(and yours) and be done with it.

spiffworkflow-backend/src/spiffworkflow_backend/models/process_instance_file_data.py
jasquat: My vote is to use the full digest. I don't see a reason to truncate but I can see how the full digest could be helpful. Also it'd be extra logic to rename it.

jakubgs: Fair enough. But honestly, what are we trying to optimize with the folder? But I wonder what exactly is the existence of this folder optimizing. In git I think it's simply a matter of limiting number of files per folder.

jasquat:
The database doesn't really need to be updated with the location of the file but it is a good idea to do that.

It may not be a bad idea to have a migration script that can move the files to / from the database. I don't think it'd be hard for us to write it.

Leaving the default location for files in the db is because not everyone needs this feature. This means we'll just be forcing complexity on users that don't need it. I like the migration idea more so if they find they do need to use this feature heavily in the future, they can easily make the switch.

spiffworkflow-backend/src/spiffworkflow_backend/models/process_instance_file_data.py
jasquat: I think limiting number of files in a directory is what we are trying to do as well.

burnettk (approved):

jasquat:
@madhurrya Yes, this PR adds a migration that will move the data from the database to the filesystem. This will happen on boot when the `SPIFFWORKFLOW_BACKEND_PROCESS_INSTANCE_FILE_DATA_FILESYSTEM_PATH` env var is set.

jakubgs:
I appreciate you implementing the migration, I had barely any time for anything recently.

>This means we'll just be forcing complexity on users that don't need it.

There's no complexity in a sane default that uses a folder/docker volume to store files. And storing files in a database is something a first year web development student would do.

PR #1414, opened 2024-04-17
Allow updating task data via a dictionary

jbirddog:
Adds support for a script:

```
update_task_data_with_dictionary({
    "x": 1,
    "y": 2,
    "z": [{"a": [1, 2, 3]}]
})
```

that when run results in `x`, `y` and `z` being variables in task data. Initially for the Iplicit use case.

## Summary by CodeRabbit

- **New Features**
	- Introduced a script that updates task data by creating or updating variables from a provided dictionary.


burnettk:
it's a sparse update, eh? it's not for deleting keys?

burnettk (approved):

jbirddog:
> it's a sparse update, eh? it's not for deleting keys?

Was thinking no since you don't easily have access to the entire task data dictionary itself. Guess it works more like Python's `somedict.update` method.

Issue #647, opened 2023-11-10
Compensation range field


harmeet-status:
Implement a solution which would allow handling the compensation amount range:
Position 2 input fields side by side, where one of them represents the minvalue and another one - maxvalue

To Do:
--------------
* Don't make this about side-by-side display of form components
* Build a custom component for "numeric_range" - should allow for float as well as integer values.
* Produce a value like:  **salary_range: "12,000...120,000.00"**, or more ideally **salary_range: {"min": 12,000, "max": 120,000.00}** 
* If the first case above, consider adding a companion function that can convert the string to two numeric values.
 
 related to #521

harmeet-status:
Comment from @calexh-sar in Discord:

“discussed the slider with the devs this morning and appears the best approach is to modify what we have to include being able to set the min value as well by sliding. I suspect we would want to update both the integerRange and integerRangeSteps options since using the latter might be the better approach for the salary range use case so the user does't have to spend extra time trying to hit the exact amount.
There was a question if a slider is the right choice for the salary range use case since the min and max values are likely to be known and the usefulness of a slider to tweak the range is probably not needed: https://www.smashingmagazine.com/2017/07/designing-perfect-slider/.
But if you and/or Veronika feel this is the best approach, a quick estimate without digging deeper is it will be a Small t-shirt size ticket, 1-3 days, so we can maintain the Carbon look and feel.”

harmeet-status:
Comment from @sashayar13:

I've been thinking also about just using two input fields, but these are the comments from KB and Alex “KB responded it's a cool idea, i think it's just harder than you might expect to implement and maybe on the order of 5 days.  Not sure if it would fall into a Medium (3-5 dev days) or Large (5-10 dev days) t-shirt size, but suspect the latter.”

![Image](https://github.com/sartography/spiff-arena/assets/104047049/606c7e49-0c77-435a-a159-3df10bb90bd2)

harmeet-status:
Comment from @sashayar13:

after the internal discussion with the team, we agreed that the option with 2 input fields is preferable. Can we pls schedule the dev work?

burnettk:
@harmeet-status in terms of scheduling the dev work, please go ahead and prioritize this one relative to other status tickets.

harmeet-status:
This one has been assigned to @Kayvon-Martinez

harmeet-status:
@Kayvon-Martinez can you update the status of this when you have started working on this.

sashayar13:
[...]

PR #1451, opened 2024-04-25
numeric-range-check-for-num

jasquat:
Addresses https://github.com/sartography/spiff-arena/issues/647#issuecomment-2076654675

Adds a validation to ensure the number entered in a numeric range field is a number.

## Summary by CodeRabbit

- **New Features**
	- Enhanced form validation in the Custom Form component to ensure the `min` and `max` values are valid numbers.
	- Improved input handling in the Numeric Range Field to prevent processing non-numeric values.


jbirddog (approved):
I don't know if javascript has a way to precompile a regex and declare it as a constant, or if there is a logical place to put one that these two files can reference, but just a thought.

PR #1550, opened 2024-05-13
allow large tokens to pass through nginx

burnettk:
fix client issue where a large token caused 414 Request-URI Too Large


madhurrya:
@burnettk what is meant by large token here? Is it the auth token or something else?

burnettk:
yes, auth token.

PR #1553, opened 2024-05-15
get a token object for local development

HunterIsMe:
When developing the backend locally, sometimes I wanted to debug my backend interface through postman or a similar tool, so I wrote a script to get tokens easily

## Summary by CodeRabbit

- **New Features**
  - Introduced a script to easily obtain an authentication token for local development.


jbirddog:
Thanks for the PR. We have some env vars that you can check when building the urls for the backend - https://github.com/sartography/spiff-arena/blob/main/docker-compose.yml#L24

jasquat:
Thanks for the PR.

The command `OPENID_TOKEN_URL='http://localhost:7000/openid/token' ./bin/get_token` also exists which should accomplish the samething but we may move over to using this one since it is python. We made this branch from this one so we can mess with it https://github.com/sartography/spiff-arena/tree/HunterIsMe-main.

jasquat:
We merged the scripts together in the other branch and merged with PR #1585. It defaults to keycloak but the command I pasted above will still work as is.

PR #1652, opened 2024-05-31
Add git-debranch as a dev dependency, allow running in arena dev container

jbirddog:
This gives the arena container a user and group and matches the current user and its primary group. Also optionally allows mounting you ssh directory so git commands can be run within the dev container. Adds `git-debranch` as a dev dependency which can be run from within the dev container if desired. If you do not want to mount your ssh dir you can run `make git-debranch-offline`.

## Summary by CodeRabbit

- **New Features**
  - Added new Makefile targets: `git-debranch` and `git-debranch-offline` for managing Git branches with and without internet access.
  
- **Enhancements**
  - Updated handling of user and group information in Makefile and Dockerfile for better customization and security.
  - Added `git-debranch` dependency to `pyproject.toml` for enhanced Git branch management.
  
- **Improvements**
  - Included a new volume in `dev.docker-compose.yml` for SSH directory management.
  - Dockerfile now installs `vim-tiny` alongside existing tools for improved development environment.

### DIFF
--- a/dev.docker-compose.yml
+++ b/dev.docker-compose.yml
@@ -10,3 +10,4 @@ services:
       XDG_CACHE_HOME: "/app/.cache"
     volumes:
       - ./:/app
+      - ${MY_SSH_DIR:-/tmp/__spiff_nossh}:${MY_SSH_DIR:-/tmp/__spiff_nossh}
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,9 @@ fe-sh:
 fe-unimported:
 	$(IN_FRONTEND) npx unimported
 
+git-debranch:
+	$(IN_ARENA) poetry run git-debranch
+
 poetry-i:
 	$(IN_ARENA) poetry install --no-root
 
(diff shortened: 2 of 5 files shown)


dev.docker-compose.yml
```diff
@@ -10,3 +10,4 @@ services:
       XDG_CACHE_HOME: "/app/.cache"
     volumes:
       - ./:/app
+      - ${MY_SSH_DIR:-/tmp/__spiff_nossh}:${MY_SSH_DIR:-/tmp/__spiff_nossh}
```
burnettk, line 13: This is the first time the dot SSH directory has been mounted? Does debranch need it?

burnettk (approved):

dev.docker-compose.yml
jbirddog: By default it runs a fetch and prune, which requires network access - so if you did an ssh clone it needs it. You have to opt in to having the ssh directory mounted, so if you don't want to fetch and prune you can run the offline version, or if you did an https remote i'd imagine it would work as well.

Issue #1656, opened 2024-06-03
Allow for optional Data Store connection


calexh-sar:
If a Data Store update is conditional, a error is throw if it is not used.

burnettk:
@jbirddog you fixed this, maybe?

jbirddog:
Just forgot to update the ticket. It is ready for qa

dinithihj:
This seems to be fixed now. 
Tested on dev.app (version 412/413) with the model https://dev.app.spiff.status.im/process-models/misc:qa:data-store-tests:optional-data-store-writes

calexh-sar:
@dinithihj what else is needed to move this issue to Ready for Release.

dinithihj:
> @dinithihj what else is needed to move this issue to Ready for Release.

@calexh-sar just waited for PRs to be approved, to have a quick test on test.app.

dinithihj:
Tested and verified on test.app (version 216/218) with the model https://test.app.spiff.status.im/process-models/misc:qa:data-store-tests:optional-data-store-writes

PR #1700, opened 2024-06-10
Prevent KeyError for optional data store writes

jbirddog:
This has come up a couple times during process model creation - if you have a data store wired up for a write from a script task, but inside that script task only want to write under certain conditions a KeyError would arise. Instead of forcing the bpmn authors to work around the KeyError, we can just skip trying to write if the data store variable has not been set in task data.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved data integrity by adding checks to ensure `self.bpmn_id` is present in task data before proceeding with operations, preventing potential issues in various data stores.


burnettk (approved):

jbirddog:
Work on #1656

Issue #1190, opened 2024-03-11
Dictionary changed size during iteration


burnettk:
when running certain activities (seen in mass create POs subprocess) in parallel. we may be adding to the subprocesses dictionary while running get_task_from_id.

https://sartography.sentry.io/issues/5055128964/events/3b88e11476884aaab88401a268521995/?project=4503966086529024

current investigate happening with this model: https://dev.app.spiff.status.im/process-models/misc:d:mi-test

harmeet-status:
Had a discussion with @sashayar13, this needs to be done either in Sprint 24 or Sprint 25 as we need this for Go-Live for Implicit 1st April.

harmeet-status:
I'm hoping that this fix: https://github.com/sartography/spiff-arena/issues/1006 will solve the original issue: https://github.com/sartography/spiff-arena/issues/1177

harmeet-status:
@calexh-sar are there any other ticket related to Celery issues? Can we pls link them here.

calexh-sar:
Moved above #1177 since this might be a fix for it and is needed by April 1.

burnettk:
elizabeth says she's looking to make some library changes that may fix or at least alleviate the problem.

essweine:
This PR may fix this issue (and will at least reduce the frequency of it occurring) https://github.com/sartography/SpiffWorkflow/pull/398

calexh-sar:
@MarD0607 reported this error likely occurred on prod today, this was the error:
```
Event Error Details
dictionary changed size during iteration

Stacktrace:Traceback (most recent call last):
  File "/app/src/spiffworkflow_backend/services/process_instance_queue_service.py", line 132, in dequeued
    yield
  File "/app/src/spiffworkflow_backend/services/process_instance_processor.py", line 1481, in do_engine_steps
    return self._do_engine_steps(exit_at, save, execution_strategy_name, execution_strategy)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib/python3.12/site-packages/SpiffWorkflow/bpmn/workflow.py", line 211, in get_task_from_id
    for subprocess in self.subprocesses.values():
RuntimeError: dictionary changed size during iteration
```

sashayar13:
https://dev.app.spiff.status.im/i/5804

[...]

PR #1711, opened 2024-06-11
Fix dictionary changed size during iteration error

jbirddog:
Work on #1190 

Worked with @essweine - now that the core library performance changes are in we were able to reproduce the issue 100% of the time with ~200 iterations. With this fix we have been able to run without the error 100% of the time, testing up to 600 iterations. To limit the amount of locking required we are only locking when the spiff task about to be run has a child that is a sub workflow task.

## Summary by CodeRabbit

- **New Features**
  - Improved task execution reliability by introducing conditional locking in the workflow execution process.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/services/workflow_execution_service.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/services/workflow_execution_service.py
@@ -5,17 +5,18 @@ import time
 from abc import abstractmethod
 from collections.abc import Callable
 from datetime import datetime
+from threading import Lock
 from typing import Any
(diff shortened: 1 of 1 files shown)


jasquat (approved):

burnettk (approved):

madhurrya:
@jbirddog could you please share the test model you used for this.

jbirddog:
> @jbirddog could you please share the test model you used for this.

Sure thing, it is here - https://github.com/sartography/sample-process-models/tree/jon/misc/jonjon/mi-test

madhurrya:
Thank you @jbirddog

PR #1713, opened 2024-06-11
fix-force-lock-pi-event

jasquat:
Only create the force run pi event when appropriate.

## Summary by CodeRabbit

- **New Features**
  - Introduced a conditional check for `force_run` when running process instances, allowing more control over process execution.

- **Tests**
  - Added a new test to verify the behavior of running process instances with the `force_run` option, ensuring robustness and correctness.


madhurrya:
@jasquat is there anything I can test related to this?

burnettk:
before, we would get process_instance_force_run events in the events tab even if the instance had never been force run. now that should no longer be the case.

PR #1781, opened 2024-06-20
bump lib to omit default on inclusive gateway with matched conditions

burnettk:
this brings exclusive gateway behavior in line with the bpmn spec.


madhurrya:
@burnettk is this about inclusive gate ways or exclusive gateways? Title and description are different.
Is there a way to see the relevant issue which has details about this update?

burnettk:
Inclusive, updated. This was the one Volker issue that didn’t get a ticket. I have forwarded you an email with details.

PR #1829, opened 2024-06-26
update rjsf for perf release

burnettk:
this allows us to get off our branch, which also had the perf fix.

## Summary by CodeRabbit

- **Dependencies**
  - Updated `@rjsf/core` to version `5.18.5`.
  - Updated `@rjsf/mui` to version `5.18.5`.
  - Updated `@rjsf/utils` to version `5.18.5`.
  - Updated `@rjsf/validator-ajv8` to version `5.18.5`.
  - Removed `react-jsonschema-form` dependency.


madhurrya:
@burnettk can you please explain about this update and about the perf release?

burnettk:
When typing in form fields was lagging, we fixed that a while back. But we used our own branch of rjsf. Now they have incorporated the fix into an official release, so we can use that instead. Ideally, this change has no functional impact, but if it did, that issue where typing in form fields was slow might return.

PR #1856, opened 2024-07-02
group metrics by endpoint instead of path

burnettk:
path includes parameters with high cardinalities, which is not good for prometheus metrics.

## Summary by CodeRabbit

- **Enhancements**
  - Improved monitoring setup by grouping Prometheus metrics by endpoint. This enhancement provides more granular insights into endpoint-specific performance and usage.


jasquat (approved):

jakubgs (commented):
Thank you.

Issue #1883, opened 2024-07-08
Ability to configure sqlalchemy options


StephenPCG:
I deployed spiffworkflow locally via docker, and configured to use postgres. I often encounter this error alert:

```
OperationalError: (psycopg2.OperationalError) server closed the connection unexpectedly
```

<img width="451" alt="image" src="https://github.com/sartography/spiff-arena/assets/836455/91235d20-4976-4375-9f99-a5db7445d90a">

I did some googling, and it suggested to add `pool_pre_ping` option to sqlalchemy, see [this stackoverflow link](https://stackoverflow.com/q/55457069)

I took a glance at spiffworkflow_backend source code, and seems it's not possible to configure via environ variable or configuration file. It would be great to be able to have it configurable. Or maybe there are other possible solutions?

burnettk:
i guess you modified the docker compose file at spiff-arena/docker-compose.yml for postgres? here's another version that is known to work with postgres: https://github.com/sartography/arena-compose-postgres/. does it work for you? we're happy to add configuration options if they are needed, but we've seen postgres work without this option.

StephenPCG:
I didn't notice that there was a docker compose file for postgres. 

I'm indeed deploying arena in k8s, I translated the docker compose file into k8s resources (deployment, configmap, etc.), provided an entrypoint to install `libpq5` on container startup (so I don't have to build a custom image and push to some registry).

Here is the entrypoint:

```bash
#!/bin/bash
set -e

apt-get update
DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends libpq5

exec /app/bin/boot_server_in_docker
```

Here is environment variables set on spiffworkflow-backend container:

```yaml
  SPIFFWORKFLOW_BACKEND_ENV: production # NOT SURE what values are allowed here? 
  FLASK_DEBUG: "0"
  FLASK_SESSION_SECRET_KEY: "[REDACTED]"
  SPIFFWORKFLOW_BACKEND_URL: "[REDACTED]"
  SPIFFWORKFLOW_BACKEND_BPMN_SPEC_ABSOLUTE_DIR: /app/process_models
  SPIFFWORKFLOW_BACKEND_CONNECTOR_PROXY_URL: http://spiffworkflow-connector
  SPIFFWORKFLOW_BACKEND_DATABASE_TYPE: postgres
  SPIFFWORKFLOW_BACKEND_DATABASE_URI: postgresql://USER:PASS@HOST/DB
[...]

PR #1903, opened 2024-07-11
Support config SQLAlchemy pool_pre_ping option from env (#1883)

StephenPCG:
Add support for config SQLAlchemy pool_pre_ping option from env, and defaults to be true.

This will make app more robust, when an idle connection in pool is closed form server side, the app will not checkout that connection and throw an error to end user.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new database configuration parameter `SPIFFWORKFLOW_BACKEND_DATABASE_POOL_PRE_PING` to improve connection reliability. The default setting is `True` for seamless database operations.


burnettk (approved):

burnettk:
thank you very much, @StephenPCG !

Issue #2037, opened 2024-08-07
Handling of BPMN Error Events


essweine:
BPMN Error Events should be handled like exceptions: they should propogate up the call activity/subprocess stack and if are not caught at the top level, caught by the engine.  The instance would be terminated by default, but we might offer the option of suspending the instance instead.

To get this working as expected, we would need to
* update the spiffworkflow library to handle internal upward propogation
* modify spiffarena-backend to handle events that were not caught inside the workflow

calexh-sar:
@madhurrya would you please get with Jon to discuss testing this issue.

madhurrya:
@calexh-sar Sure will do. Is it 'ready for QA'?

madhurrya:
@burnettk please let me know when this PR is deployed to somewhere I can test. https://github.com/sartography/spiff-arena/pull/2058

madhurrya:
1. Call activity related issue

This model shows the unhandled Error as expected.
https://mysql.spiffcrm.com/process-models/misc:qa:bpmn-model-testing:unit-test-call-activity-with-error-end-event

But this doesn't
https://mysql.spiffcrm.com/process-models/misc:call-activity-error-end-event:level-2
and this too doesn't
https://mysql.spiffcrm.com/process-models/misc:qa:bpmn-model-testing:unit-test-call-activity-with-error-boundary-event

I am not sure why. Is it because the first one is inside a sub process and these two are not? But I don't think the error end events has to be inside a sub process always.

madhurrya:
3. Call activity related issue

In both these models I have called the same error end event model.
This one shows the unhandled Error
https://mysql.spiffcrm.com/editor/process-models/misc:qa:bpmn-model-testing:unit-test-call-activity-with-error-end-event/files/unit-test-call-activity-with-error-end-event.bpmn

But this one doesn't
https://mysql.spiffcrm.com/editor/process-models/misc:qa:call-activity-tests:call-activity-level-2/files/call-activity-level-2.bpmn

madhurrya:
@jbirddog the above issues are fixed now. All of them show the unhandled Error now.

But noticed this behavior.

The ones that were working as expected earlier, when we run them after displaying the error those instances completes.
e.g.
https://mysql.spiffcrm.com/i/477
![image](https://github.com/user-attachments/assets/39b82ce0-c0ac-402e-8061-b431020ee942)

https://mysql.spiffcrm.com/i/480
[...]

PR #2058, opened 2024-08-26
Bubble up unhandled error and escalation events at the end of the workflow

jbirddog:
If the workflow ends with unhandled errors or escalation events, bubble them up as exceptions so they are handled in the normal flow of workflow errors. Bumps SpiffWorkflow to bring in the library support.


jasquat (approved):

madhurrya:
@jbirddog please check the issues I have added here https://github.com/sartography/spiff-arena/issues/2037

PR #2082, opened 2024-09-13
Allow for optional AZP validation in authentication_service

### LINKED ISSUE #2077: Authorized Party Check does not make sense
In the spiff backend authentication service, the "azp" of the jwt is checked to match the client id of spiff itself. See [authentication_service.py#L377](https://github.com/sartography/spiff-arena/blob/ca88178700ed96a28f29eecb1726c2c6ec58d665/spiffworkflow-backend/src/spiffworkflow_backend/services/authentication_service.py#L377)
However, this check does not make sense in case of an exchanged token. My user logs in via a different service and then my backend performs a token exchange for the spiff client in order to talk to the API. The audience check should suffice, as the authorized party may legitimately be another client.

pitwegner:
Closes #2077.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration option for enhanced control over OpenID verification settings.
  
- **Bug Fixes**
	- Improved validation logic for the authorized party (AZP) parameter in the authentication process, allowing for better flexibility based on configuration settings.


pitwegner:
Is there a place where environment/config variables need to be documented?

burnettk (approved):

burnettk:
we don't have a separate place for documenting these other than `default.py` itself. thanks for the contribution!

PR #2087, opened 2024-09-20
Feature/user model upsert

pitwegner:
When not using the frontend, users do not get created automatically, because there is no login. When simply using the backend, it would be desirable to allow user creation from the first time the token is used.


spiffworkflow-backend/src/spiffworkflow_backend/routes/authentication_controller.py
```diff
@@ -377,6 +377,8 @@ def _get_user_model_from_token(decoded_token: dict) -> UserModel | None:
                         .first()
                     )
                     if user_model is None:
+                        if current_app.config["SPIFFWORKFLOW_BACKEND_CREATE_USERMODEL_WITHOUT_LOGIN"]:
+                            user_model = AuthorizationService.create_user_from_sign_in(decoded_token)
                         AuthenticationService.set_user_has_logged_out()
                         raise ApiError(
                             error_code="invalid_user",
```
burnettk, line 384: this is a cool feature, thanks. it looks like right after it creates the user model in this case, it will return an ApiError. is it possible for the request to succeed to make it more seamless?

burnettk (approved):

PR #2090, opened 2024-09-26
Enable having created files match the local user:group

mogul:
If you run `docker compose up` with no other arguments, the `process_models` directory and everything in it will be created with `root:root` ownership. That's not great for the local editing use-case, because then you've got to use root permissions to work with those files. 

With this proposed change, running `UGID="$(id -u):$(id -g)" docker compose up` should result in any created files matching the ownership of the invoking user. If the user doesn't set `UGID` and just runs `docker compose up`, then the existing behavior (owner `root:root`) is the the default.

---

NOTE: This is currently a _draft_ PR because I haven't yet figured out how to get around [the invocation of `git config` on startup](hhttps://github.com/sartography/spiff-arena/blob/main/spiffworkflow-backend/bin/boot_server_in_docker#L89)... That command wants to write to `/root/.gitconfig`, which has `root:root` ownership in the image. 🤔

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,9 @@ services:
   spiffworkflow-backend:
     container_name: spiffworkflow-backend
     image: ghcr.io/sartography/spiffworkflow-backend:latest
+    # Enable setting the ownership of created files to match the local user:group. For example:
+    #     UGID="$(id -u):$(id -g)" docker compose up -d
+    user: ${UGID:-0:0}
     environment:
       SPIFFWORKFLOW_BACKEND_ENV: "local_docker"
       FLASK_DEBUG: "0"


mogul:
> NOTE: This is currently a _draft_ PR because I haven't yet figured out how to get around the invocation of `git config` on startup... That command wants to write to `/root/.gitconfig`, which has `root:root` ownership in the image. 🤔

What do you think of [this solution](https://github.com/sartography/spiff-arena/pull/2090/commits/6c58b3ba68583755229e959a19cf03ae12d7199d)? I don't believe this has any security implications, but I don't know everywhere that you use this image and what you might map into it.

burnettk (approved):
excellent, thanks

burnettk:
we only use the one git repository inside the container, so yeah, i don't think there are any security implications. we might end up putting the same git config statement in the main Dockerfile as well.

mogul:
> we might end up putting the same git config statement in the main Dockerfile as well.

I think we have to... This isn't working as-is:

```
2024-09-30 19:37:26 BOOT SERVER IN DOCKER (2024-10-01 02:37:26): Running git init
2024-09-30 19:37:26 Reinitialized existing Git repository in /app/process_models/.git/
2024-09-30 19:37:25 {"level": "DEBUG", "message": "SPIFFWORKFLOW_BACKEND_BPMN_SPEC_ABSOLUTE_DIR: /app/process_models", "loggerName": "spiffworkflow_backend", "processName": "MainProcess", "processID": 7, "threadName": "MainThread", "threadID": 281473289465888, "timestamp": "2024-10-01T02:37:25.546Z"}
2024-09-30 19:37:26 INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
2024-09-30 19:37:26 INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
2024-09-30 19:37:26 error: could not lock config file //.gitconfig: Permission denied
2024-09-30 19:37:26 Exited with BAD EXIT CODE '255' in ./bin/boot_server_in_docker script at line: 89.
```

PR #2096, opened 2024-10-01
Try to head off file ownership issues

mogul:
See https://github.com/sartography/spiff-arena/pull/2090#issuecomment-2384667055

I _think_ this will work, but I'm not working directly with your build pipeline locally.🤞

cc @burnettk 

## Summary by CodeRabbit

- **Chores**
	- Updated the Dockerfile to improve the build process and resolve dependency issues.
	- Configured Git to avoid file ownership issues in the deployment stage.
	- Changed Poetry version to 1.6.1 for better compatibility with PEP 517 builds.
	- Added necessary development libraries for Python dependencies.

### DIFF
--- a/spiffworkflow-backend/Dockerfile
+++ b/spiffworkflow-backend/Dockerfile
@@ -31,6 +31,9 @@ RUN apt-get update \
 RUN pip install --upgrade pip
 RUN pip install poetry==1.8.1
 
+# Avoid host:container file ownership issues
+RUN git config --global --add safe.directory *
+
 
 ######################## - SETUP
 
@@ -68,5 +71,4 @@ LABEL source="https://github.com/sartography/spiff-arena/spiffworkflow-backend"
 LABEL description="Backend component of SpiffWorkflow, a software development platform for building, running, and monitoring executable diagrams"
 
 COPY --from=setup /app /app
-
 CMD ["./bin/boot_server_in_docker"]


mogul:
Unfortunately this does not successfully enable `RUN_AS` to work, because the RUN_AS user does not have permission to create the SQLite DB on [the Docker-managed DB volume that's mounted into the backend container](https://github.com/sartography/spiff-arena/blob/341a89c2538e2835e8c9e537f95730559288d30d/docker-compose.yml#L70-L72), which has root:root permissions.😞

I'm noodling on how that can be resolved, since this seems like it must be a pretty common situation with a well-understood solution. An alternative is to introduce a Postgresql container that uses that volume, and have the backend use Postgres instead of SQLite. I think this change would better reflect actual deployment scenarios, while being incidental to the local "docker compose up" workflow we're trying to make better here. What do you think?

burnettk:
definitely true that mysql and postgres are the right targets for real deployed environments. https://github.com/sartography/arena-compose-postgres has a compose file using postgres.

mogul:
[A WILD SOLUTION APPEARS!](https://stackoverflow.com/a/73255981) This looks like The Right Way, being both clever and straightforward at the same time. It's something I can easily test locally; will do that, then make a PR if it works.

burnettk:
:D

mogul:
https://github.com/sartography/spiff-arena/pull/2099

PR #2153, opened 2024-11-21
be consistent about how the redirect uri is generated.

danfunk:


danfunk:
These changes are already in place on emerson.spiff.works

burnettk (approved):

PR #2171, opened 2024-12-03
Enable using read-only HTTPS repositories

mogul:
If someone wants to clone read-only from an `https://github.com/...` URL, that should be permitted, and in that case they shouldn't be required to provide an SSH key. (Without this, if you provide neither the path nor the key, then you see `ERR /app/bin/clone_process_models: line 39: SPIFFWORKFLOW_BACKEND_GIT_SSH_PRIVATE_KEY_PATH: unbound variable` in the logs and the clone on the penultimate line never happens.)

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of SSH private key for Git operations to prevent errors during cloning.

- **Refactor**
	- Enhanced clarity in the assignment and export process of the SSH private key path.
	- Streamlined the conditional export of the SSH command based on the availability of the private key.


spiffworkflow-backend/bin/clone_process_models
```diff
@@ -24,18 +24,23 @@ fi
 
 if [[ -z "${SPIFFWORKFLOW_BACKEND_GIT_SSH_PRIVATE_KEY_PATH:-}" ]]; then
   if [[ -n "${SPIFFWORKFLOW_BACKEND_GIT_SSH_PRIVATE_KEY:-}" ]]; then
-    export SPIFFWORKFLOW_BACKEND_GIT_SSH_PRIVATE_KEY_PATH=$(mktemp /tmp/ssh_private_key.XXXXXX)
+    SPIFFWORKFLOW_BACKEND_GIT_SSH_PRIVATE_KEY_PATH=$(mktemp /tmp/ssh_private_key.XXXXXX)
```
mogul, line 27: CodeRabbit is missing a subtlety here that shellcheck.net caught: If you export on the same line that you run a subshell, you might mask an error in the execution of the subshell. Not likely here with mktemp, but good BASH style says the export should happen on a separate line from the assignment for that reason.

burnettk (approved):

PR #2174, opened 2024-12-04
Refactor image build to use a matrix build, add security scanning

mogul:
With a little help from Claude.AI...! This consolidates multiple jobs into a single `create_docker_images` job using a matrix strategy. Along the way it removed a couple of (hopefully innocuous) inconsistencies between the three versions of the process.

That turned out well so I also added security-scanning between image build and image push. Currently it doesn't push if there are any `critical` or `high` findings; you may want to adjust that. It also doesn't do anything to notify you when that happens; not sure if there's a good way to do that, or if GitHub Security will let you know on its own. 🤔

## Summary by CodeRabbit

- **New Features**
	- Consolidated Docker image building into a single job for improved efficiency.
	- Added a vulnerability scanning step to enhance security checks.

- **Improvements**
	- Streamlined workflow for building and pushing Docker images.
	- Enhanced organization through a matrix strategy for managing multiple images.

### DIFF
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -100,6 +100,8 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           platforms: linux/amd64,linux/arm64
+          cache-from: type=gha
+          cache-to: type=gha,mode=max          
       - name: Adding markdown
         run: echo 'TAGS ${{ steps.meta.outputs.tags }}' >> "$GITHUB_STEP_SUMMARY"
 
--- a/.github/workflows/build_docker_images.yml
+++ b/.github/workflows/build_docker_images.yml
@@ -34,11 +34,24 @@ on:
     tags: [v*]
 
 jobs:
-  create_frontend_docker_image:
+  create_docker_images:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - image_name: sartography/spiffworkflow-frontend
+            context: spiffworkflow-frontend
+            description: "Frontend component of SpiffWorkflow, a software development platform for building, running, and monitoring executable diagrams"
+          - image_name: sartography/spiffworkflow-backend
+            context: spiffworkflow-backend
+            description: "Backend component of SpiffWorkflow, a software development platform for building, running, and monitoring executable diagrams"
+          - image_name: sartography/connector-proxy-demo
+            context: connector-proxy-demo
+            description: "spiffworkflow-connector-proxy-demo"
+
     env:
       REGISTRY: ghcr.io
-      IMAGE_NAME: sartography/spiffworkflow-frontend
+      IMAGE_NAME: ${{ matrix.image_name }}
       BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/build_docker_images.yml
```diff
@@ -34,67 +34,24 @@ on:
...
-      - name: Build and push Frontend Docker image
-        uses: docker/build-push-action@v6.10.0
-        with:
-          # this action doesn't seem to respect working-directory so set context
-          context: spiffworkflow-frontend
-          push: true
-          tags: ${{ steps.meta.outputs.tags }}
-          labels: ${{ steps.meta.outputs.labels }}
-          platforms: linux/amd64,linux/arm64
-      - run: echo 'TAGS' >> "$GITHUB_STEP_SUMMARY"
```
mogul, line 90: I couldn't see what this line was doing, so I took it out... Let me know if it was load-bearing and I'll put it back in!

.github/workflows/build_docker_images.yml
```diff
@@ -132,67 +89,13 @@ jobs:
             type=ref,event=tag,enable=true,format=latest
 
       - name: Write app version info
```
mogul, line 91: This step was not previously happening for the `connector-proxy-demo` image, but with this refactor it is happening, to be consistent with the other two. Let me know if that was intentionally omitted previously!

mogul:
CodeRabbit b0rked the sequence diagram in its last update: We've split the build and push into separate processes, where the build loads the images into the local Docker daemon for the action, and then only pushes to GHCR and Docker Hub after scanning and finding no critical/high problems (and uploading the scan results) in between. Let's see if it picks it up this time since I was rate-limited after my last commit...?

@coderabbit review

mogul:
Very close, @coderabbit, but the "Pull images from Docker daemon" interaction should happen after the "Scan images for vulnerabilities" interaction.

mogul:
@CodeRabbit better, thanks. However, I just realized that there's no step that pushes images to Docker Hub in this workflow, so the "Docker Hub" actor and the interactions with it can be removed from the sequence diagram. Can you also add an interaction between "GitHub Actions" and itself, labeled "Test QuickStart guide steps", representing the steps in the GitHub Action after the image-building and image-pushing is done?

mogul:
Nailed it.

mogul:
I think testing that the QuickStart steps work with the new images should also happen _**before**_ pushing the images, rather than after, which is how the existing code worked prior to this PR. I'm leaving that change out of scope for this PR unless you see a quick or obvious way to make that improvement, @burnettk (or, I guess, @coderabbit)!

mogul:
That seems about right. Can you actually make a commit that would accomplish those changes, or is it up to me @CodeRabbit?

mogul:
I'm not crazy about exporting and reimporting the images as tar files, and I see a few other problems with that which I'd want to fix. What do you think @burnettk? Fix it here, or fix it a different PR in future?

.github/workflows/build_docker_images.yml
burnettk: seems great. this sets us up to add this information into an api method in that app, which seems like it would have no downside.

.github/workflows/build_docker_images.yml
burnettk: this makes the github summary page more useful by outputting the pushed tags, badly formatted though it may be. the line was preserved at the end of jobs.create_docker_images, so i think we're good.

burnettk:
totally!

mogul:
OK, well, that didn't work very well.😞
image

mogul:
OK, here's the reason... [Loading only supports one architecture at a time.](https://github.com/docker/buildx/issues/59#issuecomment-1168619521) PR incoming.

burnettk:
@mogul after your second PR, the system was working, and the build started failing because of critical and high severity vulnerabilities in backend, frontend, and connector. 🥳 we ratcheted it down to just failing on critical level vulnerabilities, fixed the critical vulnerabilities by updating the Dockerfiles to ensure that updated packages were used, and it fixed the build (and there are corresponding closed items in the code scanning section of github). thank you very much for this contribution. this new tooling and visibility improves our security posture and gives us a strong foundation on which to iterate.

PR #2280, opened 2025-03-23
File upload process

usama9500:
Added a section on file upload process model: https://emerson.spiff.works/process-models/misc-processes:example-file-upload/files/example-file-upload.bpmn

## Summary by CodeRabbit

- **Documentation**
  - Introduced a new guide outlining the file upload process and how users can obtain a markdown download link, complete with illustrations to enhance clarity.
  - Updated the navigation to include the new file upload section, making it easier for users to access the guide.


burnettk (approved):
Looks good. Maybe fewer images next time. Those images that just show what step of the diagram you are describing seem unnecessary to me. Maybe just one image of the full diagram is sufficient.

usama9500:
> Looks good. Maybe fewer images next time. Those images that just show what step of the diagram you are describing seem unnecessary to me. Maybe just one image of the full diagram is sufficient.

Removed it.

burnettk:
i think the images are still in this change, so they would still be added to the repo, but are just not referenced in the markdown.

usama9500:
> i think the images are still in this change, so they would still be added to the repo, but are just not referenced in the markdown.

I have kept the output screens so we want to show the how upload screen will look.

burnettk (approved):

PR #2301, opened 2025-04-17
Support dynamic items in forms

jbirddog:
Adds support for pulling an array field's items from task data, much like dynamic enumerations. The value in task data is in the same format as what the rjsf forms require. If we did the same thing for the enumerations we could drop this [block](https://github.com/sartography/spiff-arena/blob/main/spiffworkflow-backend/src/spiffworkflow_backend/routes/process_api_blueprint.py#L799), but that would require diagram changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of form options by supporting both "anyOf" and "items" keys, ensuring select options from task data are correctly displayed in more scenarios.
  - Enhanced compatibility with various data formats for form options, allowing for broader support of different task data structures.


burnettk (approved):
you managed to have it still be valid jsonschema before it gets "templated"? awesome!

jbirddog:
> you managed to have it still be valid jsonschema before it gets "templated"? awesome!

I can't promise that it is but it doesn't obviously blow up or cause errors - however I doubt it could be used outside the system without our handling.

PR #2345, opened 2025-05-23
fix(user): do not fail on URL parse

richtlu:


burnettk (approved):
gonna merge and then make a follow-on change here.

PR #2346, opened 2025-05-23
fix(auth): use backend_url for redirection

richtlu:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the construction of redirect URLs during login to ensure consistent use of the configured backend URL.


burnettk:
it may be helpful to use `/v1.0/debug/url-info` to see how backend is interpreting the host, etc.

PR #2350, opened 2025-05-31
feat(translate): extend translations and add czech language

richtlu:
Implement more i18n strings using AI translation and add Czech language.
New Flag icon with language lists in SideNav.


jasquat:
@richtlu Sorry it took so long to get back to you. I seemed to have missed the update.

To answer your question, we use different function in different contexts mostly for formatting issues and sometimes so we can pass in random errors that are not in the state since ErrorDisplay uses only errors in the state and we do not always want that.

We are looking into how to fix this now.

burnettk (approved):

PR #2364, opened 2025-06-11
feat(username): record username to db instead of email

richtlu:
Not sure what is the story behind, but I am proposing to use `username` column for username field from openid provider instead of email.

The use case is to have a multiple accounts for one email.


richtlu:
@jasquat first of all it is confusing for me now; maybe I am missing something, but why is there a column which is a duplication of another one?
The use case is that people in our organization would like to use several accounts tied with one email for context sharing and permission separation. E.g. there would be a `boss`  and  `boss.assistant1` account; both report actions to the same email address, but first one has enhanced permissions and the other hasn't

jasquat (approved):

PR #2410, opened 2025-07-21
Feature/allow reset to error tasks

essweine:
This allows a process to be reset to a task that has failed, in lieu of resetting to its parent, which could result in repeating needless work or work with side-effects.

## Summary by CodeRabbit

* **New Features**
  * Added support for resetting and resuming process instances from tasks in the "ERROR" state.
  * Introduced a new BPMN process definition to test error handling and reset functionality.

* **Bug Fixes**
  * Tasks in the "ERROR" state now display the completion view link, similar to completed tasks.

* **Tests**
  * Added a unit test to verify correct process reset and resumption after encountering an error task.


burnettk (approved):
looks good, tests pass.

PR #2509, opened 2025-09-07
Alllow restricting access to spiffworkflow-frontend

burnettk:
this isn't really a security feature or impenetrable defense, but more of a policy suggestion and lock on the front door. the intent is to be able to use this in scenarios where you are expected to take actions over the API (probably via a custom frontend) and not through the spiffworkflow-frontend UI.

## Summary by CodeRabbit

- New Features
  - Frontend now gates access and shows an “Access Denied” screen if the user lacks permission.
  - Health check includes an access flag so the UI can reflect per-user availability.
  - Default permissions updated so basic/support roles can access the frontend.

- Localization
  - Added “Access Denied” title and message in EN, DE, ES, PT-BR, PT-PT, FI, CS, and ZH-CN.

- Tests
  - Simplified browser tests to use the built-in Page fixture.
  - Added dedicated test project configuration.

- Documentation
  - Added README with instructions for running browser tests.


spiffworkflow-frontend/src/ContainerForExtensions.tsx
```diff
@@ -55,7 +57,7 @@ export default function ContainerForExtensions() {
     permissionRequestData,
   );
 
-  const { removeError } = useAPIError();
+  const { removeError: _removeError } = useAPIError();
```
jasquat: Why the rename?

spiffworkflow-frontend/src/ContainerForExtensions.tsx
```diff
@@ -138,15 +140,6 @@ export default function ContainerForExtensions() {
     }
   }, [isMobile]);
 
-  // never carry an error message across to a different path
```
jasquat, line 141: was removing this useEffect intentional?

burnettk: yeah, it was a dup

spiffworkflow-frontend/src/ContainerForExtensions.tsx
burnettk: undoing, thanks.

PR #2520, opened 2025-09-10
Bring in the async http connector

jbirddog:
Starting to copy over some new-style connector proxies things. This is the async http connector that has a v2 response without the v2 suffix. Currently it only boots for the dev containers and is not deploying on merge to main. Want to shutdown the other one first.

## Summary by CodeRabbit

- New Features
  - Introduces an asynchronous HTTP connector proxy with endpoints for GET/POST/PUT/PATCH/DELETE/HEAD.
  - Supports headers, query params, bodies, and optional basic auth.
  - Provides a metadata endpoint describing available operations.
  - CORS enabled and efficient JSON handling for faster responses.

- Developer Experience
  - New local service with Docker-based dev and prod configurations, hot-reload support, and easy shell/log access via Make targets.

- Chores
  - Expanded .gitignore to exclude common local artifacts.


connector-proxies/async-http/fly.toml
```diff
@@ -0,0 +1,22 @@
+# fly.toml app configuration file generated for acp-bitter-fog-6526 on 2024-10-22T09:57:03-04:00
+#
+# See https://fly.io/docs/reference/configuration/ for information about how to use this file.
+#
+
+app = 'acp-bitter-fog-6526'
```
burnettk, line 6: probably no harm in exposing this, right?

burnettk (approved):

PR #2540, opened 2025-09-22
log started milestone when creating an instance from message start event

burnettk:
i think we noticed this because of analytics. analytics are now picking up message start events, i believe, presumably through a different change. shall we abandon this branch? cc @essweine @jasquat 

## Summary by CodeRabbit

* **New Features**
  * Added event logging when a process is started in response to a message, improving auditability and observability.
  * Logs now include milestone information plus process model and instance identifiers for message-triggered starts, making it easier to trace execution history.


jasquat (approved):

burnettk:
sweet, thanks

PR #2617, opened 2025-11-14
Support PKCE and refresh token rotation

Mwindo:
This PR accomplishes two things:

1. Allow [PKCE](https://www.rfc-editor.org/rfc/rfc7636) in the OAuth login flow, which adds a layer of security
2. Allow the user access token to be refreshed more than once. This is important for setups in which the access token is aggressively expired for security (e.g., in a stateless JWT session approach). Currently, SpiffWorkflow only updates the refresh token once even if it receives an updated refresh token in the refresh response.

Note that the fix for 2, while necessary, is not sufficient depending on the settings of the auth server. Multiple [near-]simultaneous requests can lead to SpiffWorkflow making multiple [near-]simultaneous requests to the auth server to refresh a token. If the auth server only allows the refresh token to be used once, for instance, these multiple auth-server requests can cause issues: the auth server will say, "Hey, you've already used this refresh token" and return an error response.

(Anecdotally, using `threading.Lock()` with a context manager around the call to the auth server to refresh the token fixed the issue. However, `threading.Lock()` does not coordinate distributed processes; a robust solution to this race condition issue would probably require something in the persistence layer (either a lock, a promise-like structure that all processes await, etc.). For now, we are solving the issue by configuring our auth server to revoke refresh tokens less aggressively.)

## Release Notes

* **New Features**
  * Added PKCE (Proof Key for Authorization Code Exchange) support for enhanced OAuth security with an optional enforcement configuration setting.
  * Improved refresh token persistence and reuse during authentication token exchanges.

* **Style**
  * Minor formatting adjustments to configuration and workflow files.


.github/workflows/auto-merge-dependabot-prs.yml
```diff
@@ -28,14 +28,12 @@ jobs:
     if: ${{ github.actor == 'dependabot[bot]' && github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success' }}
     steps:
       - name: Development Code
-        uses: actions/checkout@v5
-
-      ###### GET PR NUMBER
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```
Mwindo: @mogul @btylerburton Do we want these changes to the yml files, or should I scope just the PKCE/refresh token changes?

spiffworkflow-backend/src/spiffworkflow_backend/services/authentication_service.py
```diff
@@ -87,6 +88,26 @@ class AuthenticationOptionNotFoundError(Exception):
...
 
+class PKCE:
+    '''
+    Config to support PKCE (Proof Key for Code Exchange)
+    RFC: https://www.rfc-editor.org/rfc/rfc7636
+    '''
+
+    CODE_VERIFIER_KEY = "code_verifier"
+    CODE_CHALLENGE_KEY = "code_challenge"
+    CODE_CHALLENGE_METHOD_KEY = "code_challenge_method"
```
Mwindo, line 100: The PKCE flow: the `code_challenge_method` (fixed to `"S256"` in this PR for security reasons) and `code_challenge`--a hashed version of `code_verifier`--are sent to the auth server on getting an authorization code.  The `code_verifier` is sent to the auth server during the request that exchanges this authorization code for an access token. This ensures that the client that received the authorization code is the same client requesting the access token.

Mwindo: I've scoped everything to this class mostly for organizational reasons.

spiffworkflow-backend/src/spiffworkflow_backend/services/authentication_service.py
```diff
@@ -321,6 +342,16 @@ def get_login_redirect_url(self, authentication_identifier: str, final_url: str
             + f"scope={' '.join(current_app.config['SPIFFWORKFLOW_BACKEND_OPEN_ID_SCOPES'])}&"
             + f"redirect_uri={redirect_url}"
         )
+
+        if (current_app.config["SPIFFWORKFLOW_BACKEND_ENFORCE_PKCE"]):
```
Mwindo: I didn't see anyplace obvious to test this, but I am happy to add a test if I have just overlooked a good spot.

.github/workflows/auto-merge-dependabot-prs.yml
btylerburton: Let's only include the token related stuff.

spiffworkflow-backend/src/spiffworkflow_backend/routes/authentication_controller.py
```diff
@@ -368,6 +368,8 @@ def _get_user_model_from_token(decoded_token: dict) -> UserModel | None:
                     if user:
                         refresh_token = AuthenticationService.get_refresh_token(user.id)
                         if refresh_token:
+                            # TODO: This can lead to race conditions depending on how strictly refresh tokens are revoked
```
btylerburton: is this sill a TODO? and if so, should the comment be revised?

Mwindo: This TODO is still active as a robust fix to this would be trickier. (This is the "not sufficient" part of the PR description.) I can change the comment, but I don't know what to change it to.

Mwindo: (The small tweak in this PR at least allows us to refresh our refresh token so long as we don't revoke their reuse too strictly.)

Mwindo: Maybe I could say, "Note: ..." or "Warning: ..."

mogul: You might substantially quote your description from the PR here as documentation of the situation for people working on this code in future, who won't know to look for this PR.

PR #2619, opened 2025-11-17
Support PKCE and refresh token rotation

Mwindo:
This PR accomplishes two things:

1. Allows [PKCE](https://www.rfc-editor.org/rfc/rfc7636) in the OAuth login flow, which adds a layer of security
2. Allows the user access token to be refreshed more than once. This is important for setups in which the access token is aggressively expired for security (e.g., in a stateless JWT session approach). Currently, SpiffWorkflow only updates the refresh token once even if it receives an updated refresh token in the refresh response.

Note that the fix for 2, while necessary, is not sufficient depending on the settings of the auth server. Multiple [near-]simultaneous requests can lead to SpiffWorkflow making multiple [near-]simultaneous requests to the auth server to refresh a token. If the auth server only allows the refresh token to be used once, for instance, these multiple auth-server requests can cause issues: the auth server will say, "Hey, you've already used this refresh token" and return an error response.

(Anecdotally, using `threading.Lock()` with a context manager around the call to the auth server to refresh the token fixed the issue. However, `threading.Lock()` does not coordinate distributed processes; a robust solution to this race condition issue would probably require something in the persistence layer (either a lock, a promise-like structure that all processes await, etc.). For now, we are solving the issue by configuring our auth server to revoke refresh tokens less aggressively.)

## Summary by CodeRabbit

* **New Features**
  * Optional OAuth PKCE support with enforceable PKCE mode and one-time verifier handling.
  * Maintenance command to purge expired PKCE verifiers.

* **Improvements**
  * Login state now encodes richer payloads and includes PKCE identifiers when enabled.
  * Backend will attempt token refresh and persist refresh tokens for smoother sessions.
  * New config option to enforce PKCE.

* **Bug Fixes**
  * Clearer error responses for missing/failed PKCE data during token exchange.

* **Tests**
  * Integration tests covering PKCE flows, enforcement, refresh handling, and cleanup.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/spiffworkflow-backend/tests/spiffworkflow_backend/integration/test_authentication.py
+++ b/spiffworkflow-backend/tests/spiffworkflow_backend/integration/test_authentication.py
@@ -27,7 +27,9 @@ class TestAuthentication(BaseTest):
     def test_get_login_state_without_pkce_enabled(self, app: Flask) -> None:
         with self.app_config_mock(app, "SPIFFWORKFLOW_BACKEND_OPEN_ID_ENFORCE_PKCE", False):
             redirect_url = "http://example.com/"
-            state_payload = AuthenticationService.generate_state_payload(authentication_identifier="default", final_url=redirect_url)
+            state_payload = AuthenticationService.generate_state_payload(
+                authentication_identifier="default", final_url=redirect_url
+            )
             state = AuthenticationService.encode_state_payload(state_payload)
             state_dict = ast.literal_eval(base64.b64decode(state).decode("UTF-8"))
 
@@ -280,7 +282,7 @@ class TestAuthentication(BaseTest):
             assert params.get(PKCE.CODE_CHALLENGE_METHOD_KEY, [])[0] == "S256"
             assert isinstance(state_dict["pkce_id"], str)
 
-    def test_get_auth_token_throws_errors_for_misconfigured_pkce(self, app: Flask, mocker: MockerFixture) -> None:
+    def test_get_auth_token_throws_errors_for_misconfigured_pkce(self, app: Flask) -> None:
         with app.test_request_context(
             "/some/path",
             base_url="https://example.com/",  # this is what request.host_url will be based on


Mwindo:
I did not see an obvious place to add tests for this or to update documentation, but I could have missed something; if so, please let me know!

spiffworkflow-backend/src/spiffworkflow_backend/services/authentication_service.py
```diff
@@ -321,6 +344,17 @@ def get_login_redirect_url(self, authentication_identifier: str, final_url: str
...
+        if current_app.config["SPIFFWORKFLOW_BACKEND_ENFORCE_PKCE"]:
+            code_verifier = PKCE.generate_code_verifier()
+            # Store the verifier server-side for use when exchanging the authorization code.
+            session[PKCE.CODE_VERIFIER_KEY] = code_verifier
+            code_challenge = PKCE.generate_code_challenge(code_verifier)
+            code_challenge_method = "S256"  # Restrict PKCE challenge method to "S256" since "plain" is less secure.
+            login_redirect_url += (
+                f"&{PKCE.CODE_CHALLENGE_KEY}={code_challenge}&{PKCE.CODE_CHALLENGE_METHOD_KEY}={code_challenge_method}"
+            )
+
```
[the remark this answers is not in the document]
jasquat: @Mwindo we will put in a separate pr to change session data over to use the database to avoid this issue.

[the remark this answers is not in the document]
Mwindo: @jasquat Thank you! I thought about this distributed issue, saw flask session configuration, thought "Spiff must be using the Flask session already somewhere," and now checked that no, not really. Whoops 🤦 .

If it is easier to make a separate model (like the existing refresh token model), that makes sense as well. (The only issue here is making sure that rows don't accumulate on error. I suppose in practice this wouldn't be too common: it could be a single row per user, deleted on any non-error PKCE flow, and overwritten the next time if it happened not to be deleted.) I'm happy to set any of this up as this is, after all, a mistake from my faulty assumption above!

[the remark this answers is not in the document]
jasquat: @Mwindo yeah that might be better. We haven't been using session data for anything.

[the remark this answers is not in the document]
Mwindo: Ok. I can set that up if you'd like, unless you would rather do so--whatever is easier.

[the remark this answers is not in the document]
burnettk: yeah, if you're comfortable creating a new model and migration for this specific purpose, that would be great. also happy to pair on it if you'd like. there was not really any great place to put general openid documentation, so we created a new page, https://github.com/sartography/spiff-arena/pull/2624.

spiffworkflow-backend/src/spiffworkflow_backend/routes/authentication_controller.py
```diff
@@ -219,13 +228,25 @@ def login_api(authentication_identifier: str) -> Response:
...
 
 def login_api_return(code: str, state: str, session_state: str) -> str:
-    # state_dict = ast.literal_eval(base64.b64decode(state).decode("utf-8"))
-    # state_dict["final_url"]
+    # Temporarily logging for debug purposes
+    current_app.logger.debug('login_api_return')
+    current_app.logger.debug('code', code)
+    current_app.logger.debug('state', state)
+
+    state_dict = ast.literal_eval(base64.b64decode(state).decode("utf-8"))
```
Mwindo: @burnettk I have things working locally with PKCE and the `login_return` endpoint, and I was going to do some work here in the `login_api_return` endpoint as well ... but I am not sure I understand the current implementation, which looks like [this](https://github.com/sartography/spiff-arena/blob/b67fcff0215a5eaa98393ebeca2cc0809af8b61e/spiffworkflow-backend/src/spiffworkflow_backend/routes/authentication_controller.py#L221C1-L231C24):

```
def login_api_return(code: str, state: str, session_state: str) -> str:
    # state_dict = ast.literal_eval(base64.b64decode(state).decode("utf-8"))
    # state_dict["final_url"]

    redirect_path = f"{current_app.config['SPIFFWORKFLOW_BACKEND_API_PATH_PREFIX']}/login_api_return"
    auth_token_object = AuthenticationService().get_auth_token_object(code, redirect_path)
    access_token: str = auth_token_object["access_token"]
    if access_token is None:
        raise MissingAccessTokenError("Cannot find the access token for the request")

    return access_token
```

I haven't followed this logic _too_ far down, but it seems incorrect. E.g., we are passing in `redirect_path` for the `authentication_identifier` argument in `get_auth_token_object`. Is there something I am missing? Is this endpoint maybe deprecated? Apologies if I am missing something obvious!

jasquat: @Mwindo Yeah that endpoint is old. We will remove it. Thanks for pointing it out.

Mwindo: @jasquat Got it--thank you! I will ignore it so it doesn't clutter up the diff here. Then I will test a little more and clean things up.

spiffworkflow-backend/src/spiffworkflow_backend/services/authentication_service.py
```diff
@@ -286,19 +350,33 @@ def logout(self, id_token: str, authentication_identifier: str, redirect_url: st
...
         return redirect(request_url)
 
+    class StatePayload(TypedDict):
+        final_url: str | None
+        authentication_identifier: str
+        pkce_id: NotRequired[str]
+
     @staticmethod
-    def generate_state(authentication_identifier: str, final_url: str | None = None) -> bytes:
+    def generate_state_payload(authentication_identifier: str, final_url: str | None = None) -> StatePayload:
```
Mwindo, line 378: The reason for this refactor:

We need some way to store a one-time PKCE code verifier associated with a given user's login flow. On initial login, we don't necessarily have a user id, and we don't have, e.g., a flask session to leverage. As a workaround:
- We generate a unique id and place it in the state payload here.
- We get that  unique id from the state payload in `get_login_redirect_url` below and create a corresponding entry in our PKCE code verifier table. (Hence why `generate_state` is now `generate_state_payload` and returns a `dict`.)
- The `login_return` code (in `authentication_controller.py`) parses out that id from the state and uses it to get the code verifier from our table.
- We pass the code_verifier  in as part of the "give me a token, auth server" request (`get_auth_token_object`).

.github/workflows/build_docker_images.yml
```diff
@@ -159,5 +159,4 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
       - name: Adding markdown
-        run: echo 'TAGS ${{ steps.meta.outputs.tags }}' >> "$GITHUB_STEP_SUMMARY"
-
+        run: echo 'TAGS ${{ steps.meta.outputs.tags }}' >> "$GITHUB_STEP_SUMMARY"
```
Mwindo, line 162: Not sure why this whitespace is happening, I've tried to fix it a couple times 😕

spiffworkflow-backend/migrations/versions/e102044e51f8_.py
```diff
@@ -0,0 +1,35 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('pkce_code_verifier',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('pkce_id', sa.String(length=512), nullable=False),
+    sa.Column('code_verifier', sa.String(length=512), nullable=False),
+    sa.Column('created_at_in_seconds', sa.Integer(), nullable=False),
```
Mwindo, line 25: We could have a script to clear out old records, and maybe even a cron job ... but I am not sure if that is necessary for this PR?

burnettk: if it's approximately one record per login, i'd say no strict need to worry about it urgently.

spiffworkflow-backend/src/spiffworkflow_backend/models/pkce_code_verifier.py
```diff
@@ -0,0 +1,20 @@
...
+    The next request ("give me an access token") needs to send a one-time code verifier based on that challenge.
+    (This ensure the client that requested the auth code is the same one requesting the access token with that auth code.)
+    We store such code verifiers here.
+    """
+
+    __tablename__ = "pkce_code_verifier"
+
+    id: int = db.Column(db.Integer, primary_key=True)
+    pkce_id: str = db.Column(db.String(512), nullable=False, unique=True)  # Unique per-login PKCE lookup key
+    code_verifier: str = db.Column(db.String(512), nullable=False)
```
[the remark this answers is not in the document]
Mwindo: We could probably do something like:

```
    def delete_outdated_pkce_code_verifiers(expiration_time_in_seconds: int = 600) -> None:
        assert expiration_time_in_seconds >= 0, "expiration_time_in_seconds must be non-negative"
        now = round(time.time())
        expired_time_cutoff = now - expiration_time_in_seconds
        deleted = PkceCodeVerifierModel.query.filter(
            PkceCodeVerifierModel.created_at_in_seconds < expired_time_cutoff
        ).delete()
        db.session.commit()
        return deleted
```

But I think that might be best for a followup. The odds of a lot of these accruing quickly is low.

spiffworkflow-backend/migrations/versions/e102044e51f8_.py
Mwindo: Yep, once per login.

spiffworkflow-backend/src/spiffworkflow_backend/models/pkce_code_verifier.py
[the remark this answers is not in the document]
Mwindo: Sure, that works for me.

spiffworkflow-backend/src/spiffworkflow_backend/services/authentication_service.py
```diff
@@ -286,19 +351,33 @@ def logout(self, id_token: str, authentication_identifier: str, redirect_url: st
...
+            "authentication_identifier": authentication_identifier,
+        }
+        if current_app.config["SPIFFWORKFLOW_BACKEND_ENFORCE_PKCE"]:
+            pkce_id = secrets.token_urlsafe(32)  # Associate a unique PKCE id with the request for cross-reference later
+            state_payload["pkce_id"] = pkce_id
+        return state_payload
+
+    @staticmethod
+    def encode_state_payload(state_payload: StatePayload) -> bytes:
+        return base64.b64encode(bytes(str(state_payload), "UTF-8"))
```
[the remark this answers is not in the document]
Mwindo: This is reasonable, but the current implementation in Spiff is `str`, and I did not want to alter it.

Mwindo:
Alright, open "for real." Happy to make changes/additions as needed!

spiffworkflow-backend/migrations/versions/e102044e51f8_.py
Mwindo: On the other hand, this is probably pretty trivial. If there is a "go-to" way to place/implement a script that commits a db change, I can just call something like [this](https://github.com/sartography/spiff-arena/pull/2619#discussion_r2547181639) and call it a day, leaving it up to dev to decide when to run it.

burnettk: sure, feel free to add a `system_maintenance` directory or similar under `bin` and shove a python script in there. thank you.

spiffworkflow-backend/src/spiffworkflow_backend/models/pkce_code_verifier.py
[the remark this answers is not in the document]
Mwindo: I think this issue can be closed. I implemented a script to delete the records, and I think it should be up to the devs using the software to determine how/when/if to run it.

spiffworkflow-backend/migrations/versions/e102044e51f8_.py
Mwindo: Done! Thanks for your patience/guidance as I get acclimated to the project structure.

spiffworkflow-backend/src/spiffworkflow_backend/config/default.py
```diff
@@ -130,6 +130,7 @@ def config_from_env(variable_name: str, *, default: str | bool | int | None = No
 config_from_env("SPIFFWORKFLOW_BACKEND_OPEN_ID_LEEWAY", default=5)
 config_from_env("SPIFFWORKFLOW_BACKEND_OPEN_ID_INTERNAL_URL_IS_VALID_ISSUER", default=False)
 config_from_env("SPIFFWORKFLOW_BACKEND_OPEN_ID_SCOPES", default="openid,profile,email")
+config_from_env("SPIFFWORKFLOW_BACKEND_ENFORCE_PKCE", default=False)  # Set to enforce OAuth PKCE (recommended)
```
burnettk: i guess it's not strictly speaking openid, but is it close enough to openid that we could include it within the group of config vars with the SPIFFWORKFLOW_BACKEND_OPEN_ID prefix do you think?

spiffworkflow-backend/migrations/versions/e8fccf29fe24_.py
```diff
@@ -0,0 +1,35 @@
...
+Create Date: 2025-11-20 12:45:21.142802
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'e8fccf29fe24'
+down_revision = 'efa7a9f771a6'
```
jasquat: This conflicts with migrations in main. I think setting this to `88188256d1a5` should be enough to resolve the conflict.

spiffworkflow-backend/src/spiffworkflow_backend/models/pkce_code_verifier.py
```diff
@@ -0,0 +1,23 @@
...
+
+    # In case there are accumulated entries, use created_at_in_seconds to determine outdated entries
+    created_at_in_seconds: int = db.Column(db.Integer, nullable=False)
```
burnettk: can we add `index=True` here for the cleanup job's benefit?

spiffworkflow-backend/src/spiffworkflow_backend/config/default.py
Mwindo: Sure, I don't think that's too much of a stretch--good suggestion!

spiffworkflow-backend/src/spiffworkflow_backend/models/pkce_code_verifier.py
Mwindo: 100%

spiffworkflow-backend/migrations/versions/e8fccf29fe24_.py
Mwindo: Got it, thank you!

Mwindo: I needed to rerun the migration anyway, so this should be fixed.

spiffworkflow-backend/src/spiffworkflow_backend/services/authentication_service.py
```diff
@@ -1,12 +1,15 @@
...
 import enum
 import json
+import secrets
 import sys
 import time
 from hashlib import sha256
 from hmac import HMAC
 from hmac import compare_digest
 from typing import Any
+from typing import NotRequired
```
jasquat: See https://github.com/sartography/spiff-arena/blob/main/spiffworkflow-backend/src/spiffworkflow_backend/interfaces.py#L11 for defining this to support python 3.10.

spiffworkflow-backend/src/spiffworkflow_backend/services/authentication_service.py
```diff
@@ -87,6 +98,78 @@ class AuthenticationOptionNotFoundError(Exception):
...
+        if code_verifier_object:
+            code_verifier_object.code_verifier = code_verifier
+        else:
+            code_verifier_object = PkceCodeVerifierModel(pkce_id=pkce_id, code_verifier=code_verifier)
+        db.session.add(code_verifier_object)
+        try:
+            db.session.commit()
+        except Exception as e:
+            db.session.rollback()
+            raise PkceCodeVerifierStorageError(f"We could not store the PKCE code verifier. Original error is {e}") from e
```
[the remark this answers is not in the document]
Mwindo: This is done automatically by SpiffworkflowBaseDBModel here: https://github.com/sartography/spiff-arena/blob/4e61b9919c2e86a8bc76a21a9df537e7da82ea56/spiffworkflow-backend/src/spiffworkflow_backend/models/db.py#L78-L83

jasquat (approved):

PR #2676, opened 2026-01-12
Save draft task data per task guid

Mwindo:


Mwindo:
The idea here is sound, but the branch is messy. I'll reopen with a cleaner branch (and a better migration so that saved task data is not wiped).

PR #2705, opened 2026-02-27
Add script to get users assigned to a task

Mwindo:
This PR adds a script for getting potential task owners, which is useful for, e.g., seeing everyone to whom a task is currently "assigned."

This helps us with a use case we have: admins with the ability to add a user to a task need to be able to see that assignment reflected in our UI.

### DIFF
--- a/spiffworkflow-backend/tests/spiffworkflow_backend/scripts/test_get_users_assigned_to_task.py
+++ b/spiffworkflow-backend/tests/spiffworkflow_backend/scripts/test_get_users_assigned_to_task.py
@@ -80,5 +80,5 @@ class TestGetUsersAssignedToTask(BaseTest):
             process_model_identifier="test_process_model",
         )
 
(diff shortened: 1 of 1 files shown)


spiffworkflow-backend/src/spiffworkflow_backend/scripts/get_users_assigned_to_task.py
```diff
@@ -0,0 +1,40 @@
...
+from spiffworkflow_backend.models.human_task_user import HumanTaskUserModel
+from spiffworkflow_backend.models.script_attributes_context import ScriptAttributesContext
+from spiffworkflow_backend.models.user import UserModel
+from spiffworkflow_backend.scripts.script import Script
+
+
+class GetUsersAssignedToTask(Script):
+    @staticmethod
+    def requires_privileged_permissions() -> bool:
+        return False
```
Mwindo, line 16: I think this probably is ok?

jasquat (approved):

PR #2713, opened 2026-03-03
Show all potential owners in tasks_list_my_tasks

Mwindo:
The `tasks/{process_instance_id}` endpoint invokes the function `task_list_my_tasks` in the tasks controller. It seems like the `potential_owner_usernames` is incorrect in this function: it always collapses to the user calling the function.

This PR represents one possible fix for the issue.


spiffworkflow-backend/src/spiffworkflow_backend/routes/tasks_controller.py
```diff
@@ -89,27 +90,31 @@ def task_allows_guest(
...
-            process_initiator_user,
-            process_initiator_user.id == ProcessInstanceModel.process_initiator_id,
-        )
-        .join(
-            HumanTaskUserModel,
-            HumanTaskUserModel.human_task_id == HumanTaskModel.id,
-        )
-        .filter(HumanTaskUserModel.user_id == principal.user_id)
-        .outerjoin(assigned_user, assigned_user.id == HumanTaskUserModel.user_id)
-        .filter(HumanTaskModel.completed == False)  # noqa: E712
```
Mwindo, line 111: The [tasks API](https://github.com/sartography/spiff-arena/blob/b63a0a1494902e89aa266c73aabe80d4acdc4955/spiffworkflow-backend/src/spiffworkflow_backend/api.yml#L2130-L2153) routes to this function. At this point in the function, the filters seem to collapse the potential owners to the current user rather than all possible assignees.

@burnettk @jasquat I am not sure, but I suspect this is unintentional?

Mwindo: Alternatively, maybe the API should be pointing to `task_list_for_me` instead of `task_list_my_tasks`?

Mwindo: (Thanks @btylerburton for the above question!)

jasquat: @Mwindo I think you are right. It seems like it was intended to be the full list of all users that could complete the task. I think `task_list_my_tasks` is the correct endpoint. I think `task_list_for_me` was for something else and will probably get deprecated.

Mwindo: Ok, cool. Yeah, I was just looking at `task_list_for_me` and came to a similar conclusion. I'll go ahead and add a test and revise my tentative changes in this PR.

spiffworkflow-backend/tests/spiffworkflow_backend/integration/test_authentication.py
```diff
@@ -325,8 +325,8 @@ def test_delete_expired_pkce_verifiers(
         # On creation, SpiffworkflowBaseDBModel automatically sets created_at_in_seconds to "now."
         # We override the timestamps to control expiry behavior explicitly.
         now = round(time.time())
-        expired.created_at_in_seconds = now - max_pkce_verifier_time_in_seconds
-        non_expired.created_at_in_seconds = now - max_pkce_verifier_time_in_seconds + 1
+        expired.created_at_in_seconds = now - max_pkce_verifier_time_in_seconds - 5
+        non_expired.created_at_in_seconds = now - max_pkce_verifier_time_in_seconds + 5
```
Mwindo, line 329: PKCE test that tested expiration right on the boundary [led to some flakiness](https://github.com/sartography/spiff-arena/actions/runs/22644804815/job/65629950532), so I added a bit of a guard.

burnettk: thank you very much.

burnettk:
thanks!

PR #2716, opened 2026-03-04
Add an API endpoint for getting users in a group

Mwindo:
This PR adds an API endpoint to get user information for users within a specified group. In the future, this functionality could probably be wrapped into a more general user search endpoint than the one that currently exists.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/api.yml
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/api.yml
@@ -4160,6 +4160,8 @@ components:
       properties:
         id:
           type: string
+        username:
+          type: string
         email:
           type: string
         display_name:


spiffworkflow-backend/src/spiffworkflow_backend/api.yml
```diff
@@ -4137,22 +4158,16 @@ components:
           type: array
     User:
       properties:
-        uid:
```
Mwindo, line 4140: As far as I can tell, this schema was outdated: these properties were added three years ago, they do not match the user model, and they do not appear when I search for them in the project.

Mwindo:
I was able to use [extensions](https://spiff-arena.readthedocs.io/en/latest/how_to_guides/extend_the_system/extensions.html) via the `/extensions/{modified_process_model_identifier}` endpoint to accomplish what I needed.

PR #2754, opened 2026-04-10
Feature/messages in ed

danfunk:
This allows overriding behavior for Message tasks so that we can better mock out messages when building processes in Ed.


jbirddog (approved):
Would recommend bumping the patch version so it will deploy to pypi on merge.

PR #2764, opened 2026-04-22
Fix service task delegate error

Mwindo:
In the `call_connector` method of `ServiceTaskDelegate`, [there is an attempt to log headers](https://github.com/sartography/spiff-arena/blob/1731038258e04bf020074d8d1d71d09084c3164c/spiffworkflow-backend/src/spiffworkflow_backend/services/service_task_delegate.py#L292) from an object that is either a `Response` or an `HttpConnectorResponse`. However, an object of type `HttpConnectorResponse` does not have the `headers` attribute, so this causes Service Connector calls in the `HttpConnectorResponse` execution branch to fail.

This PR fixes the issue by adding headers to the `HttpConnectorResponse` response object and using better typing hints so that such errors will be exposed in the future.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/connectors/http_connector.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/connectors/http_connector.py
@@ -1,4 +1,5 @@
 import json
+from dataclasses import dataclass
 from typing import Any
 from typing import cast
 from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


spiffworkflow-backend/src/spiffworkflow_backend/connectors/http_connector.py
```diff
@@ -9,16 +10,18 @@
 from spiffworkflow_backend.config import CONNECTOR_PROXY_COMMAND_TIMEOUT
 
 
+@dataclass
 class HttpConnectorResponse:
```
Mwindo, line 14: As far as I could tell, the only place this class is used currently is in this file, and these attributes are never `None`. I thought a dataclass was a good way to force devs to supply the required attributes.

Mwindo:
I believe the failing safety tests are unrelated to this code.

burnettk (approved):

spiffworkflow-backend/src/spiffworkflow_backend/connectors/http_connector.py
```diff
@@ -109,11 +112,10 @@ def _connector_response(http_response: requests.Response) -> HttpConnectorRespon
         "spiff__logs": [],
     }
 
-    # create a blank object so we can mimic an actual http response object
-    obj = HttpConnectorResponse()
-    obj.text = json.dumps(return_dict)
-    obj.status_code = http_response.status_code
-    return obj
+    # return a mimicked version an actual http response object
```
Mwindo, line 115: Aw shucks small typo. Oh well.

burnettk:
thanks! re: safety, https://github.com/sartography/spiff-arena/pull/2766

PR #2767, opened 2026-04-23
Fix get_current_task_data script

danfunk:
get_current_task_data was broken due to changes in the serialzer, for  this case, we just bypass the serializer and use Spiff's default registry to convert the task data to a dictionary.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/scripts/get_current_task_data.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/scripts/get_current_task_data.py
@@ -1,6 +1,6 @@
 from typing import Any
 
-from SpiffWorkflow.bpmn.serializer import DefaultRegistry
+from SpiffWorkflow.bpmn.serializer import DefaultRegistry  # type: ignore
 
 from spiffworkflow_backend.models.script_attributes_context import ScriptAttributesContext
 from spiffworkflow_backend.scripts.script import Script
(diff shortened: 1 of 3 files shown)


burnettk (approved):

spiffworkflow-backend/src/spiffworkflow_backend/scripts/get_current_task_data.py
```diff
@@ -17,5 +20,8 @@ def get_description(self) -> str:
         """
 
     def run(self, script_attributes_context: ScriptAttributesContext, *_args: Any, **kwargs: Any) -> Any:
-        task_dict = BpmnProcessService.serializer.to_dict(script_attributes_context.task)
-        return task_dict["data"]
+        spiff_task = script_attributes_context.task
+        if not spiff_task:
+            return {}
+        data = DefaultRegistry().convert(spiff_task.data)
+        return {k: v for k, v in data.items() if k not in _INTERNAL_KEYS}
```
Mwindo, line 27: Just for my own learning: is there any danger here in bypassing the `BpmnProcessService` serializer for the default?

danfunk: Its a good question.  I think it may be safer to bypass it.  The default registry is converting things to dictionaries, but the serializer is designed to go farther, and we are looking at ways to override the serializer to do more compaction, diffing, and zipping values to keep things more compact when running these workflows in memory - such as we do with the new editor.

PR #2794, opened 2026-05-05
Patch update KKV data stores

btylerburton:
This PR allow users to interact more fully with the KKV data store.

1. Adds an API for RESTful PUT access to a data store using a top_level_key and secondary_key. It also adds a way to clear those data stores individually.
2. Adds selective filtering to the GET method for data store items.
3. (FINALLY, AND SOMEWHAT CONTROVERSIALLY) Normalizes the data structure for fetching a single entry or all entries by removing the nesting of the bulk fetch within a `data:` object.

I added tests of both features, and also ended up cleaning up formatting in the api.yml as I modified it, which adds most of the changes.

Let me know if Point 3 is a deal-breaker or if you want to discuss. IMO it's much easier to code around an endpoint that returns the same signature regardless of size but I'm cautious of potentially introducing a breaking change for other consumers.

### DIFF
--- a/spiffworkflow-backend/src/spiffworkflow_backend/routes/data_store_controller.py
+++ b/spiffworkflow-backend/src/spiffworkflow_backend/routes/data_store_controller.py
@@ -107,7 +107,7 @@ def data_store_item_list(
     top_level_key = top_level_key or None
     secondary_key = secondary_key or None
 
-    if top_level_key is not None and data_store_type == "kkv":
+    if (top_level_key is not None or secondary_key is not None) and data_store_type == "kkv":
         return _kkv_filtered_items(identifier, location, top_level_key, secondary_key, page, per_page)
 
     data_store_class, _ = DATA_STORES[data_store_type]
@@ -117,12 +117,12 @@ def data_store_item_list(
 def _kkv_filtered_items(
     identifier: str,
     location: str | None,
-    top_level_key: str,
+    top_level_key: str | None,
     secondary_key: str | None,
     page: int,
     per_page: int,
 ) -> flask.wrappers.Response:
-    """Return KKV entries filtered by top_level_key and optionally secondary_key."""
+    """Return KKV entries filtered by top_level_key and/or secondary_key."""
     store_model = db.session.query(KKVDataStoreModel).filter_by(identifier=identifier, location=location).first()
 
     if store_model is None:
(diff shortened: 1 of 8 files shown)


spiffworkflow-backend/src/spiffworkflow_backend/routes/data_store_controller.py
```diff
@@ -23,7 +27,10 @@
```
Mwindo, line 32: While we are in this file, is it worth replacing the above DATA_STORES definition with:

```
type DataStoreClass = type[DataStoreCRUD]

DATA_STORES: dict[str, tuple[DataStoreClass, str]] = {
    "json": (JSONDataStore, "JSON Data Store"),
    "kkv": (KKVDataStore, "Keyed Key-Value Data Store"),
    "typeahead": (TypeaheadDataStore, "Typeahead Data Store"),
}
```

And then specifying:

```
def _build_response(
    data_store_class: DataStoreClass,
```

spiffworkflow-backend/src/spiffworkflow_backend/routes/data_store_controller.py
```diff
@@ -57,44 +64,150 @@ def data_store_types() -> flask.wrappers.Response:
...
-        results.append(result)
-    response_json = {
-        "results": results,
-        "pagination": {
-            "count": len(data.items),
-            "total": data.total,
-            "pages": data.pages,
-        },
-    }
+    if per_page == 0:
```
Mwindo: This feels unintuitive to me for two reasons:

- As an API user, I would not expect per_page=0 to return all results, especially since that could potentially return an unexpectedly large response with performance implications.
- As a developer, I would be caught off guard by the deviation from what appears to be standard pagination behavior via .paginate(...).

btylerburton: This change is not strictly necessary, and still feels incomplete. Thanks for flagging. 

I was looking for a convention to force the download of an entire data store, since that is what the current API supports, but also to standardize the signature for both filtered and unfiltered calls. There are a few options.

- In the past, I've introduced a `paginate=False` query param for this use case
- We can also leave the unfiltered call as is, and allow it to pass into the response builder.
- Or just pipe it all through the pagination route

Mwindo: > In the past, I've introduced a paginate=False query param for this use case

👍 I think this is clearer while still supporting the old behavior: we can keep the default behavior the same (`paginate = False`) and support the new pagination only when `paginate == True`.

danfunk:
Tested against or example processes models and I don't see any issues.  Checking with the rest of the team but this looks good to me.

danfunk (approved):
looks good to me.

PR #2799, opened 2026-05-06
perf: index process-group permission lookups

jbirddog:
Before: get_process_groups_user_has_permissions_to re-scanned permission assignments for each group and then re-scanned denied parent ids, which could degrade toward O(G*P + G^2) across large trees. In the regression test, matcher calls were 7600. After: precompute permit/deny indexes once, then do per-group set/prefix lookups and recursion, reducing the hot path to roughly O(P + G*depth). In the same test, matcher calls are bounded below 500.


burnettk (approved):
it's a bit hard for me to read but 70 lines of the added lines diff is tests, so i don't think it's really a lot more code. probably mostly moving stuff around. let's rock it.

jbirddog:
Cool if it causes issues we can revert np.

PR #2833, opened 2026-05-26
Message Start Events in Ed 

danfunk:
Sets up a Custom Start Event that will allow us to handle start event messages in the same we handle receive messages when executing in Ed.  It will present a form based on the message schema, and allow tests to run against the process.

## Summary by CodeRabbit

* **New Features**
  * Improved workflow handling for BPMN start events.
  * Non-message start events can now proceed automatically without additional input.
  * Message-based start events continue to pause until their message payload is provided.

* **Bug Fixes**
  * Improved start-event parsing, execution, and serialization for more consistent workflow behavior.

* **Chores**
  * Updated the common package version to 0.1.41.


jbirddog (commented):
Would regen uv.lock like coderabbit suggests else the next pr will have to do it

PR #2838, opened 2026-05-28
feat: add Process Instances column sorting

photon-cat:
## Summary
- Adds sortable Process Instances table headers backed by report metadata order_by.
- Adds an id tiebreaker so paginated sorts are deterministic.
- Leaves non-sortable derived columns as plain headers.

## Test plan
- [x] npm run check-ci


jasquat (approved):
I made a couple changes but looks good overall.

PR #2924, opened 2026-07-20
use action more conventionally

burnettk:


burnettk:
@copilot how does this look?

PR #2958, opened 2026-08-09
Test perf 2

burnettk:


spiffworkflow-backend/src/spiffworkflow_backend/services/spec_file_service.py
```diff
@@ -135,7 +134,6 @@ def get_references_for_file_contents(
     def validate_bpmn_xml(cls, file_name: str, binary_data: bytes) -> None:
         file_type = FileSystemService.file_type(file_name)
         if file_type.value == FileType.bpmn.value:
-            BpmnValidator()
```
burnettk, line 138: wasn't doing anything, as there was no call to validate().

PR #2961, opened 2026-08-12
local dev: default frontend url to http://localhost:$((PORT + 1)) instead of hardcoded 7001

photon-cat:
Running the backend on a non-default port broke login with `InvalidRedirectUrlError`, because `SPIFFWORKFLOW_BACKEND_URL_FOR_FRONTEND` stayed hardcoded to `http://localhost:7001` while the frontend conventionally runs on backend port + 1. This change derives the frontend port number from the backend port number.


jasquat:
If you are running without defaults, why not just set `SPIFFWORKFLOW_BACKEND_URL_FOR_FRONTEND` to what you need?

photon-cat:
Yes its just one less environment variable you have to set, just a nice to have thats unnecessary. also on macos port 7000 is used by default

PR #2963, opened 2026-08-12
pagination: jump to first/last page

photon-cat:
For PGE: "Now I can jump only one page forward or backward, it could be nice to have a possibility to jump to specific page or at least to the beginning or the end. I've tried to change page parameter in the URL, but after my edit it is always set to 1."*

-Added goto first page and goto last page buttons (showFirstButton / showLastButton on TablePagination)
-Small refactor extracted setPageSearchParam() since both the arrows and the dropdown now
-remove 2 page option from paginationForTable, ProcessModelTabs and TaskListTable


spiffworkflow-frontend/src/components/PaginationForTable.tsx
```diff
@@ -53,28 +62,69 @@ export default function PaginationForTable({
...
+    // native select options stay fast even with the MAX_PAGES worst case.
+    // each option shows the page number and the item range it covers, so the
+    // meaning of a page is clear regardless of the per-page setting.
+    const pageOptions = [];
+    for (let pageOption = 1; pageOption <= totalPages; pageOption += 1) {
+      const from = (pageOption - 1) * perPage + 1;
+      const to = Math.min(pageOption * perPage, totalItems);
+      pageOptions.push(
+        <option key={pageOption} value={pageOption}>
+          {`${pageOption} (${from}-${to})`}
```
jasquat: I think this may be duplicating the other range of instances. I think I'd be for removing the instance range from here in favor of the other since it feels cluttered here.

Image

spiffworkflow-frontend/src/components/PaginationForTable.tsx
```diff
@@ -27,20 +33,23 @@ export default function PaginationForTable({
 }: OwnProps) {
   const { t } = useTranslation();
   const PER_PAGE_OPTIONS = [2, 10, 50, 100];
+  const MAX_PAGES = 1000;
```
jasquat: What happens when we go over this amount? I'm not sure how we should fix that. Maybe use a `...` with the final page after it?

jasquat (changes requested):

photon-cat:
Screenshot 2026-08-13 at 10 13 32 AM

jasquat (approved):

PR #2979, opened 2026-08-21
Add index on single_process_hash for BPMN process lookups

aboutrajan:
### Summary

Add a non-unique B-tree index on bpmn_process_definition.single_process_hash to optimize the BPMN process definition lookup performed by the runtime.

The runtime executes queries of the form:

```
SELECT id, single_process_hash
FROM bpmn_process_definition
WHERE single_process_hash = '<hash>'
LIMIT 1;
```

The existing process_hash_unique index is composite and has full_process_model_hash as its leading column. As a result, it cannot be used for an index seek when filtering only on single_process_hash. MariaDB instead performs an index scan across the composite index.

On deployments with thousands of BPMN process definitions, this resulted in significantly increased lookup latency and, in our environment, API requests taking up to ~2 minutes.

### Changes

**Add the following index:**

```
CREATE INDEX ix_bpmn_process_definition_single_process_hash
  ON bpmn_process_definition (single_process_hash);
```

The index is intentionally non-unique because the existing process_hash_unique constraint continues to enforce uniqueness of the (full_process_model_hash, single_process_hash) pair.

Expected query plan

**Before:**

```
type: index
key: process_hash_unique
rows: ~N
Extra: Using where; Using index
```

**After:**

```
type: ref
key: ix_bpmn_process_definition_single_process_hash
rows: 1
Extra: Using where; Using index
```

This changes the lookup from an index scan over the table's process definitions to a direct index seek on single_process_hash. The query is also covered by the new index, so no additional table-row lookup is required for the selected columns.

### Why this belongs in the base schema

This lookup is part of the normal BPMN runtime path rather than a rare or optional operation. The performance issue can remain hidden on small development databases but becomes increasingly expensive as deployments accumulate BPMN versions and drafts.

Adding the index is a purely additive schema change with no data migration and preserves the existing uniqueness semantics.

### Verification

- Added the index through an Alembic migration.
- Verified the query plan changes from an index scan to an index seek.
- Verified that the lookup uses the new index and estimates a single matching row.
- Verified BPMN parsing/runtime requests with representative process definitions.
- Related deployment observation


aboutrajan:
Hi @burnettk  could you move this fixes to v2.4.2  version? Thanks in advance

PR #2994, opened 2026-08-30
Persist Content Studio provenance and add localization foundations

calexh-sar:
## Summary
Adds the Arena-side foundation for safely matching localized content to an exact Filestore snapshot and pinning that source identity when a process instance is created.

## What changed
- Validates Filestore file hashes and canonical manifest digests before importing a referenced snapshot. Legacy packages without the new reference remain supported.
- Persists `source_artifact_ref` on imported process models and copies it onto new process instances so later source updates cannot silently change an existing instance.
- Adds locale-catalog validation for exact source, process-model, locale, and integrity compatibility.
- Adds localization-before-Jinja helpers and protects canonical JSON Schema choice values from translation.
- Preserves queued instruction templates, task data, BPMN identity, source reference, and Git fallback revision so automated instructions can be rendered in the selected locale later.
- Adds migrations for process-instance provenance and queued-instruction render context.

## Verification
- 25 focused unit and integration tests passed using SQLite.
- Ruff and mypy checks passed.
- Both migrations applied successfully; `d6f2a9b47c31` is the single Alembic head.

## Notes
- This establishes runtime contracts and persistence; catalog publication, binding, and the Arena locale selector remain later phases.
- No workflow BPMN or diagram/DI content changed.

### DIFF
--- a/spiffworkflow-backend/migrations/versions/c8d7e4a91f2b_add_source_artifact_ref_to_process_instance.py
+++ b/spiffworkflow-backend/migrations/versions/c8d7e4a91f2b_add_source_artifact_ref_to_process_instance.py
@@ -1,0 +1,25 @@
+"""add source artifact reference to process instance
+
+Revision ID: c8d7e4a91f2b
+Revises: 2d68edd689b9
+Create Date: 2026-08-30 09:05:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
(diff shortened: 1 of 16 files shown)


calexh-sar:
Superseded by #2995, which reuses Arena existing process-model and Git-revision linkage and removes both migrations plus all process-model, process-instance, import, and queued-instruction persistence changes.

